Issue #26, opened 2016-12-22
Build Script Errors when compiling on Mono


dogwith1eye:
I'm running into few build script errors in the build.fsx file when compiling on Mono version 4.6.2

**Extra quote at end of build.fsx**

```
bash-3.2$ sudo ./build.fsx
build.fsx(76,42): error FS0514: End of file in string begun at or before here
```

Which is easily fixed by removing the extraneous quote at the end of the file

exec fakeExe ("makefile.fsx"::passedArgs)"
------------------------------------------^

**TypeInitializationException in applyExecutionPermissionUnix**

```
let applyExecutionPermissionUnix path =
    let _,stat = Syscall.lstat(path)
    Syscall.chmod(path, FilePermissions.S_IXUSR ||| stat.st_mode) |> ignore

System.TypeInitializationException: The type initializer for 'Mono.Unix.Native.Syscall' threw an exception
```

Commenting out the applyExecutionPermissionUnix function lets me build fine, which makes me wonder whether it's really a necessary step.

gmpl:
Thanks for testing it in Mono.
You can submit a PR to correct that file.
Alternatively we can re-create the solution from scratch using the [F# Project Scaffold](http://fsprojects.github.io/ProjectScaffold/) which include build scripts for all platforms.

dogwith1eye:
I think it makes more sense to re-create with the the f# project scaffolding. I can submit a PR for that if you like.

gmpl:
It would be great if you can submit that PR.
Then we will have paket, docs and tests.
That would be very useful because we will be merging FsControl here soon, which has already unit tests and some docs.

gmpl:
I guess this is fixed now by https://github.com/gmpl/FSharpPlus/commit/2fb9bcdc13b93fefed23614eb9b049d78c914d23 otherwise please re-open.

PR #27, opened 2016-12-23
Initial commit of rebuilt project

dogwith1eye:
Rebuild of project using [Project Scaffold](http://fsprojects.github.io/ProjectScaffold/)

**Notes**

I have left the `<OtherFlags>--warnon:1182</OtherFlags>` node in the fsproj file which results in a number of new warnings.

Building the documentation results in a number of error messages seemingly related to tpetricek/FSharp.Formatting#364


gmpl:
No worries, we'll have to address all the unused variables, but that's a good thing, those warnings already helped me to [spot a mistake](https://github.com/gmpl/FSharpPlus/commit/2a589bdbc68c8f5a5a7936ea90d5a2d5d5f5b072).

We'll see how can we workaround the problem with the docs.

It will be interesting to use the RC F# 4.1 in the build, but not sure if that's possible or if we have to wait until the final release.

Thanks !

gmpl:
I guess this fixes #26 otherwise please re-open.

I just realized that the history for the existing files is lost, at least here in Github, it's a shame but not sure if there's a way to workaround this.

gmpl:
I guess this will fix #26 otherwise please re-open.

I just realized that the history for the existing files is lost, at least here in Github, it's a shame but not sure if there's a way to workaround this.

dogwith1eye:
I'll see what i can find about the missing history. Will let you know.

gmpl:
Thanks. I've found [this](http://stackoverflow.com/questions/1094269/whats-the-purpose-of-git-mv) but then I'm even more confused about why it doesn't track the files.

dogwith1eye:
Yeah and it's also talked about [here](http://stackoverflow.com/questions/1094269/whats-the-purpose-of-git-mv). I ended up using [this script](https://gist.github.com/emiller/6769886) from the discussion and [forcing a push](http://stackoverflow.com/questions/5509543/how-do-i-properly-force-a-git-push)

I have submitted another PR using this method and the history is preserved for the moved files. The other option is to simply keep FSharpPlus as the src folder instead of moving it under the src directory.

gmpl:
Thanks a lot.

I tried accepting it and forcing the history but it ended up with all commits duplicated.
Maybe we should not fight back the way git is designed and leave it as it is.

In principle I don't mind rewriting the history since I don't think there are many clones out there and if there are, they will have a hard time merging in any case, but now I'm thinking that also comments in the commits will stop showing and external links to specific points in the history will stop working.

Also there might be more 'moves' in the future as well.

dogwith1eye:
When you say leave as is to you want to "leave as it is", should we

-  remove the src/FSharpPlus folder and update the scaffolding to point to the original FSharpPlus source folder, this will keep all history at the expense of a very slightly different scaffolding

-  remove the original FSharpPlus folder keep the src/FSharpPlus folder, this looses the history but keeps a standard scaffolding layout

Let me know and i'll submit another PR

gmpl:
I was referring to your second point.
I think that's the way github expects you to work when you move a folder, you remove it and then create it in a different folder, or use git mv but then the rest it's up to the way you track the history.
That's my humble understanding but I'm learning Git on the fly, so I might be wrong.

I think in your original PR you didn't remove the original folder and I didn't realize it was duplicated.
Now I'm finishing integrating FsControl here, so now we can either go back in time and do the scaffold again or do a clean-up now.

Or, a 3rd option would be to go back in time and integrate FsControl first, then do the scaffolding for the project already integrated, without the FsControl dependencies.

What do you think is the cleanest way to go?

dogwith1eye:
Why don't we go with option 3 and you can finish up integrating FsControl. When that gets finish we can can rescaffold the project and simply git mv the source files to their new location. 

No messing around with rewriting history and keeping a standard layout for the library moving forward.

gmpl:
I agree with you.
Actually what I can do is go back, right before the scaffolding, and then integrate FsControl.
Now I think I have a better idea of how to integrate them properly.
Then we do the scaffolding as you suggest, for the projects already integrated.

gmpl:
I think the integration of both projects is pretty much done.
The old samples from FsControl are now in a specific folder, but those more than sample files are rather like tests files, if something goes wrong they stop compiling. It would be interesting to make the build script to discover them and compile them but we can do that in another stage.

So now if you want to scaffold it you can submit another PR.
I used ``git mv``all the time for the integration and github managed to track the file moves, [see this](https://github.com/gmpl/FSharpPlus/commit/1277304b94669cbb51829b454e56e555e5c36fa3).

The test project would have to be adapted to use NUnit, that's not very complicated but don't worry about it, I can do it afterwards.

Just in case, note that the BaseLib is statically linked into the FSharpPlus dll.

Thanks for your valuable contribution.

gmpl:
@dogwith1eye 
In case you forked it already.
I just amended the latest commit.
Hopefully it doesn't cause too much trouble for you to get it again.

dogwith1eye:
I'll take a look, i'm working on the project right now

PR #29, opened 2016-12-27
Scaffold

dogwith1eye:
Got bogged down trying to convert existing .*proj files, so instead simply moved the files and updated the .*proj files generated by Project Scaffold.

I converted the tests to NUnit as this was easier than keeping them in MSTest. Converting them back should be trivial.

Take a look and let me know if you want me to change anything.


gmpl:
Awesome !!!

This is way better than before, good job !

Unfortunately I will be travelling today with limited internet connection, but will finish reviewing later or tomorrow morning and will do the merge.

But from what I can see now it looks fine, and Github is able to track precisely every single change.

gmpl:
I noticed that the nuget files are gone now. Do you know if is this expected in this new project model?
I'm wondering what's the best way to handle versions, right now we're on pre-releases.

dogwith1eye:
Project Scaffold uses paket which is package manager that wraps nuget. You can read about it [here](https://fsprojects.github.io/Paket/getting-started.html)

To add a new package you update the paket.dependencies file in the root and paket.references file in the project root. 

```
Windows
$ .paket/paket.exe update
Linux/Osx
$ mono .paket/paket.exe update
```

That will update the packages in the .paket folder in the root.

The build.fsx file gets the release version from the RELEASE_NOTES.md file and generates a corresponding AssemblyInfo.fs file for the build.

gmpl:
So, the best way is to use the RELEASE_NOTES?
I didn't know that detail.
I was happy with myget CI000xx number that was created automatically, without need to commit anything special before publishing.
But I guess this is better to track source code of a specific binary.

gmpl:
Hmm, there is another problem.
I can't use versions like 0.9.x because we already have in nuget versions like 1.0.0-CI00037 so will have to keep using 1.0.0-CIxxxxx until the first official release.

dogwith1eye:
Yeah, i was a bit confused at first where the version was coming from. [Here](http://fsharp.github.io/FAKE/apidocs/fake-releasenoteshelper.html) is the fake release notes helper that documents the behavior.

if your replace #### 1.0.4 - December 26 2016 with #### 1.0.0-CI00004 - December 26 2016 in the Release_Notes.md file then you should be good.

I'm guessing the purpose of the release notes helper is to ensure you're keeping documentation up to date when releasing new version of the library.

Of course, there's no rule saying you have to do it this way, this was just how it's setup in the default scaffold project.

dogwith1eye:
I had copied the ItemGroup section over from the original fsproj file and forgot to take it out. When i look at the fsproj file that Project Scaffold generates, it doesn't have it in the fsproj at all.

```
<ItemGroup>
    <Compile Include="Library.fs" />
    <None Include="Script.fsx" />
    <None Include="paket.references" />
    <None Include="paket.template" />
    <Content Include="App.config" />
  </ItemGroup>
```

So I'm guessing you can just leave it out all together.

gmpl:
Yes, I also noticed that is not references on the project.
Will ignore it.
Thanks for all the information !

PR #47, opened 2017-10-20
Fixing (hopefully!) Bind signature

wasphub:
Not sure if I had to put it as `Bind` or as `(>>=)`, let's see what you say :)


wasphub:
Added requested changes, please check

PR #48, opened 2017-11-07
+netstandard and .net45

wallymathieu:
- [x] fix compilation issues related to Lazy
- [x] verify that tests are Green
- [x] make sure that build.fsx works with .net core
- [ ] make sure that travis can package nuget


gusty:
Very nice to have this PR merged.
What's the status?
I can see that travis build is failing but I think is not related to this, it started failing a while ago and I think the fix is to [upgrade FAKE](https://github.com/fsharp/FAKE/issues/1581).

wallymathieu:
I can build it on windows and get a test failure

gusty:
@wallymathieu If it's a single test that it's failing just comment it out.
Why? Because since you're not doing any code change, the only reason why a test would fail is because of a bug in the FSharp Compiler, believe it won't be the first time.
So just comment out the test stating in which situations it started failing and we'll address it somewhere else.

wallymathieu:
I haven't gotten it to compile on travis yet, so I might need to fix that.

gusty:
Well, as I said, I think it fails on Travis because of the paket version. If you can fix it that will be great but if it's too much work we can address it in another PR.
At the moment I'm happy having at least myget compiling.

wallymathieu:
Well, if you can checkout my sources and see that it builds on your machine that would be great, since then we could merge this and create a separate PR.

gusty:
Oh, it fails on my machine (Windows):

image

wallymathieu:
Do you have .net core 2 installed?

gusty:
Yes, that might be the problem, let me check if it is and eventually install it and try it again.

gusty:
I just installed it, but still getting the same.
See:
```
**********************************************************************
** Visual Studio 2017 Developer Command Prompt v15.4.2
** Copyright (c) 2017 Microsoft Corporation
**********************************************************************

C:\Users\gus.leon\source>dotnet --info
.NET Command Line Tools (2.0.2)

Product Information:
 Version:            2.0.2
 Commit SHA-1 hash:  a04b4bf512

Runtime Environment:
 OS Name:     Windows
 OS Version:  10.0.15063
 OS Platform: Windows
 RID:         win10-x64
 Base Path:   C:\Program Files\dotnet\sdk\2.0.2\

Microsoft .NET Core Shared Framework Host

  Version  : 2.0.0
  Build    : e8b8861ac7faf042c87a5c2f9f2d04c98b69f28d
```

Do I need to restart my computer?

gusty:
This is what I'm getting now:

image

wallymathieu:
Humz, that's weird. I'll see if I can reproduce it.

wallymathieu:
It might be that it needs to run `dotnet restore`. I've added it to the build script.

gusty:
I see that's building now. Will have a look at the test that fails, but shouldn't be an issue.

gusty:
Cool, both build. I can see there is an issue with the whole test projects, but in travis it works fine. On the other hand in travis there is a problem building the nuget.

wallymathieu:
yes, the main reason why it fails looks like it is due to it targeting both net45 and .netstandard2.0, might be something related to https://github.com/fsprojects/Paket/issues/2496

gusty:
I think multi targeting unit tests it's useful to discover F# compiler bugs, so definitely there is a value on multi-targeting, but I like to have this merged, since you fixed many other issues. So, worst case if we can't make it work can we remove multi-targeting in the unit tests?
Regarding the other issue, having the travis build working would be great, but if gets too complicated we can temporary disable the nuget packaging for Travis, anyway I use the myget build to publish them.

wallymathieu:
Looks like there is some issues with pack, might be only on mac and linux. Otherwise we can use "dotnet pack" to do multi-targeting of libraries on windows.

wallymathieu:
Do you use appveyor as well?

gusty:
No, I never used it. Do you think we can add it? If so I'm happy to have it.

wallymathieu:
It shouldn't be a problem to add it, you have build.cmd and appveyor yml in the root. I think that they have .net core support on their images (since it's only windows). Needs to be setup by someone with write access to the repository :wink:

gusty:
I sent you an invitation as collaborator so you can have write access.

wallymathieu:
I've started a build on appveyor: https://ci.appveyor.com/project/wallymathieu/fsharpplus/build/0.0.1.1
I'll see if it works with this branch as well later on

gusty:
Cool !

I think these kind of errors:
```
C:\projects\fsharpplus\src\FSharpPlus\Operators.fs(89,36): error FS0010: Unexpected identifier in type constraint. Expected infix operator, quote symbol or other token. [C:\projects\fsharpplus\src\FSharpPlus\FSharpPlus.fsproj]
```

makes me think it's not compiling with F# 4.1 since that syntax was not available before.

gusty:
Actually, that's seems to be the case:

     C:\Program Files (x86)\Microsoft SDKs\F#\4.0\Framework\v4.0\fsc.exe -o:obj\Release\FSharpPlus.dll --debug:pdbonly --noframework --define:TRACE --doc:bin\Release\FSharpPlus.xml --optimize+ -

wallymathieu:
I think I remember adding 
`FSharp.Compiler.Tools` in order to always build with a specific version of the fsharp compiler

gusty:
Anyway I wonder why it uses F# 4.0 by default, it's been sometime already since F# 4.1 was released.
Myget adopted it after a few weeks. Travis took some more time.

gusty:
All green in my computer !

wallymathieu:
So we should merge it. For some reason when I added appveyor it came under my account: https://ci.appveyor.com/project/wallymathieu/fsharpplus So perhaps you need to add it in order for appveyor to build pull requests?

gusty:
Does it shows green on your side?

wallymathieu:
It works on my side as well!

gusty:
Ok, I will merge.

gusty:
I went to myget and I saw that it builds perfectly and creates the nugets for 4.5 and netcore 2.0
For some reason Travis fails on creating the nugets, but that's not really an issue since I publish from MyGet, so I think we can disable the nuget step for Travis.
That way it shows green on the main Readme

wallymathieu:
Probably for the best to avoid building it on travis if everything else works. 
I also just realised that this looks wrong: https://github.com/gusty/FSharpPlus/blob/master/build.fsx#L168, the package might have a weird version.

gusty:
Yes, I had a look and myget created the 1.0.0-CI00021 but in nuget the latest version is 1.0.0-CI00089, so definitely something is wrong there. I think this was working fine some time ago, I have to check in the history when (and why) it was changed.

wallymathieu:
I think it's this pull request. https://github.com/gusty/FSharpPlus/blob/789e6540429be24ad85c98c98be98890a300ae95/build.fsx#L159-L165 is the original. I changed it while trying to get it to build on travis.

wallymathieu:
I've started working on another pull request to fix that: #49

wallymathieu:
also, if you build using "msbuild", then it works on travis as well

gusty:
Cool, thanks !

PR #59, opened 2017-12-30
Shouldn't it use the default name for map?

wallymathieu:
as I understand `fmap` is part of Haskell compatibility


gusty:
Yes. It seems to be a copy-paste error.
Thanks ;)

PR #63, opened 2018-01-02
Samples project

wallymathieu:
- This is to compile the samples, should fix part of https://github.com/gusty/FSharpPlus/issues/56
- This should be cleaner than https://github.com/gusty/FSharpPlus/pull/62#issuecomment-354683044 since it's using standard build infrastructure to compile


gusty:
Works on my machine !

wallymathieu:
```
  <Compile Include="..\FSharpPlus\Samples\*.fsx">
       <Link>Samples\%(FileName)%(Extension)</Link>
   </Compile>
```
😉

wallymathieu:
I agree that this pull request should be squashed, to reduce noise

gusty:
Nice hack !

Anyway don't delete the other branch, I still think it's easier to extend to the docs.

wallymathieu:
I think it's a standard msbuild feature that you can use in a csproj or fsproj ... so probably less hacky than the other solution. I've restored that branch for future reuse.

PR #64, opened 2018-01-04
Documentation fixes

wallymathieu:


gusty:
Looks fine.
Can you include these fixes as well?

 - The alternative in the diagram has a zero, while in the docs it's empty.
 - traverse is missing Traverse

Then I squash all as a doc fixing commit.

gusty:
Also, there are some double quotes here:

        static member Traverse (t:'Traversable<'T>, f : 'T->'``Functor<'U>) : 'Functor<'Traversable<'U>>
                                                             ^^
Since those fragments are not 'runnable' we should take them out.

wallymathieu:
How do you render the plantuml diagram?

gusty:
Yes, we have to apply the fix there as well.

Unfortunately I do it 'by hand': I paste the code in http://www.plantuml.com/plantuml/uml click submit and then I copy the generated svg URL.

I don't know if there's a better way to do that.

I also noticed that the diagram is being rendered over the navigation bar, maybe one fix is to re-position that bar on the left side for the whole doc site.

wallymathieu:
It looks like a css issue.

gusty:
We can address it in a separated PR.

I will accept this squashed as one for now.

gusty:
Here:

    <object data="//www.plantuml.com/plantuml/png/nLXBRziu
                                              ^^^

It should be svg instead of png.

I've noticed that the URL looks long and very similar to the previous one.
Maybe they are just compressing the text in the plantuml file, if we can find the compression they're using then we can automate this.

gusty:
Indeed they do: http://plantuml.com/pte

wallymathieu:
I asked someone more fluent in css than me about it and he said to use flexbox instead: https://css-tricks.com/snippets/css/a-guide-to-flexbox/

gusty:
We can give it a try.
If we’re not happy we can move the bar to the left.

gusty:
I managed to generate the compressed string for the plantuml with F#.
Now the problem is simply I can't find out how to emit it inside the URL of the object.
I tried:
 - Generating the entire object with F#, doesn't work because the text get encoded
 - Adding a key value pair to the replaces list in the call to ``Literate.ProcessDirectory``, doesn't work. Maybe it only takes into account the F# code?
 - Other ugly hacks, but they didn't worked

Can't be so complicated, maybe it's because it's too late.

PR #69, opened 2018-01-12
upload comparsion table for Haskell

ShalokShalom:
if have no idea if this is near to sane, so please check :) ;)


wallymathieu:
Does it make sense to have it there or should there be an Haskell compatibility page so that one can expand on that?

ShalokShalom:
Yeah, I also thought about this. 

@gusty?

gusty:
I agree, we should initially put it in the Haskell Compatibility doc.

Having it in the landing page will give the wrong impression that this is a Haskell emulator project, which is far from reality I hope.

Otherwise, the table looks good !

gusty:
I just realized that there is no Haskell Compatibility docs at all.
So you'll have to create one, and include it in the project, and the links, and so on.
So either:

 - You create it and put the table there.
 - I create it and you PR against it.
 - I accept you PR as it is, create the page and move the table there.

ShalokShalom:
I already create a new one.

gusty:
Great !
Let us know when you think you're ready to merge it.

ShalokShalom:
Now?

Its already uploaded. What is about the links and so on?

Nice to work with you guys ^-^

gusty:
Thanks !

I will create the links, no worries.

gusty:
Sorry for the confusion, it's uploaded now.
See the link in the readme.

PR #81, opened 2018-02-04
Example usage of F#+ Cont

wallymathieu:
- Ported [examples](https://gist.github.com/wallymathieu/ad8fe3d5b85875cf6f487ad0180be660) to F#+

### DIFF
--- a/src/FSharpPlus/Samples/Cont.fsx
+++ b/src/FSharpPlus/Samples/Cont.fsx
@@ -1,0 +1,235 @@
+#if INTERACTIVE
+#r @"../bin/Release/net45/FSharpPlus.dll"
+#else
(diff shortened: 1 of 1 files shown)


wallymathieu:
What do you think @gusty ? I think the f#+ variant looks similar to the extcore one, but nicer due to the Cont.run method in order to switch back to non cont context.

gusty:
I'm glad that you like the ``Cont.run``, the difference is of course, because in order to be able to generalize, we're using here the Haskell approach, which means wrapping and unwrapping, so we 'mark' the types instead of the operations.

I also find it nice, but it's true that when stacking many monads it becomes a bit cumbersome, although it avoids a looooot of boilerplate code.

PR #82, opened 2018-02-05
Wrapper types as structs

gusty:
Operations like ``bind`` make heavy use of wrapping-unwrapping. In order to avoid additional pressure on the GC we can set all wrapper types to structs.

### DIFF
--- a/src/FSharpPlus/Cont.fs
+++ b/src/FSharpPlus/Cont.fs
@@ -3,6 +3,7 @@
 /// <summary> Computation type: Computations which can be interrupted and resumed.
 /// <para/>   Binding strategy: Binding a function to a monadic value creates a new continuation which uses the function as the continuation of the monadic computation.
 /// <para/>   Useful for: Complex control structures, error handling, and creating co-routines.</summary>
+[<Struct>]
 type Cont<'r,'t> = Cont of (('t->'r)->'r)
 
 /// Basic operations on Cont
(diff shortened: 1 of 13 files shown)


src/FSharpPlus/Cont.fs
```diff
@@ -3,6 +3,7 @@
 /// <summary> Computation type: Computations which can be interrupted and resumed.
 /// <para/>   Binding strategy: Binding a function to a monadic value creates a new continuation which uses the function as the continuation of the monadic computation.
 /// <para/>   Useful for: Complex control structures, error handling, and creating co-routines.</summary>
+[<Struct>]
```
wallymathieu: Cont have `NoEquality; NoComparison` ?

wallymathieu: Since it wraps a function type?

gusty: Yes, but I think it doesn't require it explicitly. We can certainly added it in anycase.

wallymathieu: Usually f# compiler emits warnings when I do similar type declarations... you know the compiler better :wink:

gusty: Not really. I know better some specific parts only ;) 
But yes, it started to emit some warnings about it, for some reason I'm still not sure (have to read again the warning message) it doesn't emit it for ``Cont`` and some other wrappers. I should add to the roadmap something like checking what other attributes are still missing (ie: CompiledName, Representation, and so on)

PR #87, opened 2018-03-11
DList from fsharpx

wallymathieu:
- [x] Port tests
- [x] Cons pattern?


src/FSharpPlus/DList.fs
```diff
@@ -1,44 +1,151 @@
...
+            match l with
+            | Nil       -> finish lefts xs
+            | Unit x    -> finish lefts <| f x xs
+            | Join(x,y) -> walk (x::lefts) y xs
+        and finish lefts xs =
+            match lefts with
+            | []    -> xs
+            | t::ts -> walk ts t xs
+        in walk [] l.dc state
+    //static member Fold (x, f, z, [<Optional>]_impl) = fold f x z
```
gusty: Remove this commented out code.

src/FSharpPlus/DList.fs
```diff
@@ -1,44 +1,151 @@
...
     static member ToSeq  x = toSeq  x
     static member ToList x = toList x
     static member OfSeq  x = ofSeq  x
     static member OfList x = ofList x
     static member Fold (x, f, z, [<Optional>]_impl) = fold f x z    
 
-    static member Return x = DList (fun r -> x::r)
+    static member Return x = DList (1, x)
     static member Map (x, f) = map f x
     static member (<*>) (f, x) = ap f x
```
gusty: Although you opened the module DList you can write these operations qualified (ie DList.map instead of map), it will makes thing clear.

src/FSharpPlus/DList.fs
```diff
@@ -1,44 +1,151 @@
...
-    static member get_Empty = DList id
-    static member Append (DList x, DList y) = DList (x << y)
+    static member get_Empty = DList( 0, Nil)
+    static member Append (x:DList<_>, y:DList<_>) = DList.append x y
     
     static member ToSeq  x = toSeq  x
     static member ToList x = toList x
     static member OfSeq  x = ofSeq  x
     static member OfList x = ofList x
     static member Fold (x, f, z, [<Optional>]_impl) = fold f x z
```
gusty: You can safely remove the 4th argument, the optional one.

wallymathieu:
Should cons pattern be ported? Since it looks more like operator definition that should be in core f#+ and then be applicable for enumerables?

gusty:
You're right.
I wouldn't define a generic `cons` for now.
I can't find real world uses case for it, but if you think it could makes sense, open an issue and we can start evaluating it.
So let's remove it for now here.

gusty:
Thanks !

PR #90, opened 2018-03-12
Update generate.fsx

ShalokShalom:
There is no docs directory and also no generated html (anymore)


wallymathieu:
The failure `1) Test Failure : FSharpPlus.Tests.Parsing.parse` is due to the assumption about current timezone.

gusty:
@ShalokShalom Actually that comment is not wrong. If the builds succeed the docs folder is created and the html files are generated there.

Then if you release them, they are uploaded to the gh-pages branch.

ShalokShalom:
Oh I see

wallymathieu:
I'm going to close the pull request. Thanks @ShalokShalom and @gusty !

PR #93, opened 2018-03-13
[WIP] Fix parse DateTime

wallymathieu:
- Setting .net core to be latest : This will hopefully give the same error on travis as on Appveyor


wallymathieu:
Turns out that DateTime behavior is more consistent if you run it differently:
https://github.com/wallymathieu/TryingOutDateTime

PR #94, opened 2018-03-15
Import of DList from FSharpx.Collections

wallymathieu:
- Using a single file instead of both a DList.fs and DList.fsi.
- Adding dependency on FsCheck for the tests.


src/FSharpPlus/DList.fs
```diff
@@ -1,45 +1,260 @@
...
-
-    static member Return x = DList (fun r -> x::r)
-    static member Map (x, f) = map f x
-    static member (<*>) (f, x) = ap f x
-    static member Join x = join x
-    static member (>>=) (x, f) = bind x f
\ No newline at end of file
+    static member ToSeq  x = DList.toSeq  x
+    static member ToList x = DList.toList x
+    static member OfSeq  x = DList.ofSeq  x
```
wallymathieu: Shouldn't there be a OfList ?

src/FSharpPlus/DList.fs
```diff
@@ -1,45 +1,260 @@
...
 module DList =
-    let empty = DList id
-    let toList (DList x)   = x []
-    let toSeq  (DList x)   = x [] :> seq<_>
-    let ofList source = source |> List.append |> DList
-    let ofSeq  source = source |> Seq.toList |> List.append |> DList
-    let append (DList x) (DList y) = DList (x << y)
-    let singleton x = DList (fun r -> x::r)
-    let cons x (DList f) = (DList (fun l -> x::(f l)))
-    let snoc (DList f) x = (DList (fun l -> f (x::l)))
```
wallymathieu: `snoc` has changed name to `conj`, this is a breaking change for any user of F#+

src/FSharpPlus/DList.fs
gusty: Well, in theory it's not really required for foldables, ``ofList`` function would work if it's a foldable, which requires ``OfSeq``.

But the situation today is that ``OfList`` doesn't have defaults defined, so it won't work.

I would say leave it off for now, once I create the defaults for ``OfList`` we can add it.

src/FSharpPlus/DList.fs
gusty: Yes, it is. Anyway I was going to suggest to use ``add``.

src/FSharpPlus/DList.fs
wallymathieu: Sounds good.

src/FSharpPlus/DList.fs
wallymathieu: I'll rename it to add.

src/FSharpPlus/DList.fs
```diff
@@ -1,45 +1,260 @@
...
+
+    ///O(n). Returns a seq of the DList elements.
+    let inline toSeq (l:DList<'T>) = l :> seq<'T>
+
+    // additions to fit f#+ :
+    let inline map f (x:DList<_>)    = DList.foldBack (cons << f ) x empty
+    let concat x = DList.fold append empty x 
+    let inline join (f:DList<DList<_>>) = concat f
+    let inline ap f x = join <| map (fun y -> map ((|>) y) f) x
+    let inline bind m k              = DList.foldBack (append << k) empty m
```
gusty: This get inferred to ``(DList<'a> -> ('b -> #DList<'a>) -> DList<'a>)`` maybe we can remove the flexible type or simply make ``DList`` sealed.

src/FSharpPlus/DList.fs
```diff
@@ -1,45 +1,260 @@
...
+
+    ///O(n). Returns a list of the DList elements.
+    let inline toList l = foldBack (List.cons) l [] 
+
+    ///O(n). Returns a seq of the DList elements.
+    let inline toSeq (l:DList<'T>) = l :> seq<'T>
+
+    // additions to fit f#+ :
+    let inline map f (x:DList<_>)    = DList.foldBack (cons << f ) x empty
+    let concat x = DList.fold append empty x
```
gusty: Same here, it gets ``(DList<#DList<'b>> -> DList<'b>)``. Actually by removing the flexible type it leads to the same function as ``join``. In that case I would remove ``join`` since in most F# collections is it called ``concat``.

src/FSharpPlus/DList.fs
wallymathieu: Sounds good 👍

src/FSharpPlus/DList.fs
wallymathieu: I'll add that to the pull request.

wallymathieu:
Thanks!

PR #98, opened 2018-03-18
In order to avoid confusing intellisense

wallymathieu:
I've applied `[<EditorBrowsable(EditorBrowsableState.Never)>]` to the below types in case where you might use the explicit type name in order to operate on data.
- `DList`, `ZipList`, `ParallelArray`, `NonEmptyList`

- `ResultT`
- `ContT`
- `OptionT`
- `Reader`, `ReaderT`
- `SeqT`
- `State`, `StateT`
- `Writer`, `WriterT`

See https://github.com/gusty/FSharpPlus/issues/84#issuecomment-372952783

### DIFF
--- a/src/FSharpPlus/Monoids.fs
+++ b/src/FSharpPlus/Monoids.fs
@@ -1,11 +1,14 @@
 ﻿namespace FSharpPlus.Data
 
 open FSharpPlus.Operators
+open System.ComponentModel
 
 /// The dual of a monoid, obtained by swapping the arguments of append.
 [<Struct>]
 type Dual<'t> = Dual of 't with
+    [<EditorBrowsable(EditorBrowsableState.Never)>]
     static member inline get_Zero () = Dual (getZero ())        : Dual<'T>
+    [<EditorBrowsable(EditorBrowsableState.Never)>]
     static member inline (+) (Dual x, Dual y) = Dual (plus y x) : Dual<'T>
 
 /// Basic operations on Dual
@@ -15,7 +18,9 @@ module Dual = let run (Dual x) = x          : 'T
 /// The monoid of endomorphisms under composition.
 [<Struct; NoEquality; NoComparison>]
 type Endo<'t> = Endo of ('t -> 't) with
+    [<EditorBrowsable(EditorBrowsableState.Never)>]
     static member get_Zero () = Endo id                : Endo<'T>
+    [<EditorBrowsable(EditorBrowsableState.Never)>]
     static member (+) (Endo f, Endo g) = Endo (f << g) : Endo<'T>
 
 /// Basic operations on Endo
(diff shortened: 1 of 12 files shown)


src/FSharpPlus/Monoids.fs
```diff
@@ -1,11 +1,14 @@
...
 
 open FSharpPlus.Operators
+open System.ComponentModel
 
 /// The dual of a monoid, obtained by swapping the arguments of append.
 [<Struct>]
 type Dual<'t> = Dual of 't with
+    [<EditorBrowsable(EditorBrowsableState.Never)>]
     static member inline get_Zero () = Dual (getZero ())        : Dual<'T>
+    [<EditorBrowsable(EditorBrowsableState.Never)>]
```
gusty: I wouldn't hide operators, we're expecting them to be used, even if it's not through F#+.

src/FSharpPlus/Monoids.fs
```diff
@@ -15,7 +18,9 @@ module Dual = let run (Dual x) = x          : 'T
 /// The monoid of endomorphisms under composition.
 [<Struct; NoEquality; NoComparison>]
 type Endo<'t> = Endo of ('t -> 't) with
+    [<EditorBrowsable(EditorBrowsableState.Never)>]
```
gusty: All the ``get_`` members I would leave them visible. No specific reasoning, simple because they are visible in standard F# classes.

src/FSharpPlus/Monoids.fs
```diff
@@ -43,21 +52,29 @@ type Any = Any of bool with
 type Const<'t,'u> = Const of 't with
 
     // Monoid
+    [<EditorBrowsable(EditorBrowsableState.Never)>]
     static member inline get_Zero () = Const (getZero ()) : Const<'T,'U>
+    [<EditorBrowsable(EditorBrowsableState.Never)>]
     static member inline (+) (Const x : Const<'T,'U>, Const y : Const<'T,'U>) = Const (plus x y) : Const<'T,'U>
 
     // Functor
+    [<EditorBrowsable(EditorBrowsableState.Never)>]
```
gusty: Here ``Map`` is not really redundant.

wallymathieu:
In this case, how do we define non redundant? Is it that they have no equivalent exposed via module? I.e. that the redundant methods are only those that have the following look:
```
     static member Operation  x = T.operation  x
```
?

gusty:
Yes, that's the case. The whole purpose is to avoid having in intellisense, a list containing:

    .map f x
    .Map (x, f)

But any deviating case (like operators) is OK, this attribute is in fact our last resort.

wallymathieu:
Looks like it's mostly `Map` that has "redundant" definition.

gusty:
Yes, one of the reasons why I was proposing to replace it with the ``<!>`` operator.
It's also possible that in the past some functions were not included in order to avoid this redundance.
For example ``.Local`` which ideally should exists as ``.local`` as well.

PR #100, opened 2018-03-23
Validations 

### LINKED ISSUE #51: Validation like Either but with an accumulating Applicative
Hi!

I would like to use FSharpPlus to get a Validation data type like 
https://hackage.haskell.org/package/Validation

@gusty Would you mind giving me **some guidance** to implement basic Validation using Applicative Functor + NonEmptyList with FSharpPlus?

Notes:
1. In the long term I would like to use that in Xamarin.Android
2. I've also seen http://fsprojects.github.io/FSharpx.Extras/reference/fsharpx-validation.html, but seems to receive less attention from maintainers. Current versions are not able to target Xamarin.Android :(

wallymathieu:
This fixes #51

### DIFF
--- a/src/FSharpPlus/Validations.fs
+++ b/src/FSharpPlus/Validations.fs
@@ -96,7 +96,7 @@ type AccValidation<'err,'a> with
         AccValidation.apply f x
   // as Alternative (inherits from Applicative)
   static member inline get_Empty () = AccFailure ( getEmpty() )
-  static member inline Append (x:AccValidation<_,_>, y:AccValidation<_,_>) = AccValidation.appAccValidation Control.Append.Invoke x y
+  static member inline (<|>) (x:AccValidation<_,_>, y:AccValidation<_,_>) = AccValidation.appAccValidation Control.Append.Invoke x y
 
   // as Functor
   static member Map        (x : AccValidation<_,_>, f) = AccValidation.map f x
--- a/src/FSharpPlus/Validations.fs
+++ b/src/FSharpPlus/Validations.fs
@@ -1,0 +1,137 @@
+module FSharpPlus.Validations
+
+open FSharpPlus
+open FSharpPlus.Lens
+open FSharpPlus.Data
+
+/// An 'AccValidation' is either a value of the type 'err or 'a, similar to 'Result'. However,
+/// the 'Applicative' instance for 'AccValidation' /accumulates/ errors using a 'Semigroup' on 'err.
+/// In contrast, the Applicative for 'Result' returns only the first error.
+///
+/// A consequence of this is that 'AccValidation' is not a monad. There is no F#+ 'Bind' method since
+/// that would violate monad rules.
+///
+/// An example of typical usage can be found <https://github.com/qfpl/validation/blob/master/examples/src/Email.hs here>.
+///
+type AccValidation<'err,'a> =
+  | AccFailure of 'err
+  | AccSuccess of 'a
+
+module AccValidation=
+  let inline map (f:'T->'U)=function
+    |AccFailure e -> AccFailure e
+    |AccSuccess a -> AccSuccess (f a) 
+  let inline apply e1' e2' = 
+    match e1',e2' with
+    |AccFailure e1, AccFailure e2 -> AccFailure (plus e1 e2)
+    |AccFailure e1, AccSuccess _  -> AccFailure e1
+    |AccSuccess _, AccFailure e2 -> AccFailure e2
+    |AccSuccess f, AccSuccess a -> AccSuccess (f a)
+  let inline foldr f x = function 
+    |AccSuccess a -> f a x
+    |AccFailure _ -> x
+
+  let inline traverse f = function 
+    |AccSuccess a -> AccSuccess <!> f a
+    |AccFailure e -> result (AccFailure e)
+
+  let inline bimap f g = function
+    |AccFailure e -> AccFailure (f e)
+    |AccSuccess a -> AccSuccess (g a)
+
+  let inline bifoldr f g x = function 
+    |AccSuccess a -> g a x
+    |AccFailure e -> f e x
+
+  let inline bitraverse f g = function 
+    | AccSuccess a -> AccSuccess <!> g a
+    | AccFailure e -> AccFailure <!> f e
+
+  /// 'bind' binds through an AccValidation, which is useful for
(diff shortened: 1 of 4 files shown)


wallymathieu:
But there is still need to move documentation into samples folder and add docs. Perhaps @oscarvarto has some ideas how to document that?

src/FSharpPlus/Validations.fs
```diff
@@ -0,0 +1,137 @@
+module FSharpPlus.Validations
```
gusty: I think we should discuss the names.
My impression is that this should be in the ``.Data`` namespace.
I'm not sure if we should call it ``AccValidation`` or just ``Validation``.

src/FSharpPlus/Validations.fs
```diff
@@ -0,0 +1,137 @@
...
+  let inline map (f:'T->'U)=function
+    |AccFailure e -> AccFailure e
+    |AccSuccess a -> AccSuccess (f a) 
+  let inline apply e1' e2' = 
+    match e1',e2' with
+    |AccFailure e1, AccFailure e2 -> AccFailure (plus e1 e2)
+    |AccFailure e1, AccSuccess _  -> AccFailure e1
+    |AccSuccess _, AccFailure e2 -> AccFailure e2
+    |AccSuccess f, AccSuccess a -> AccSuccess (f a)
+  let inline foldr f x = function
```
gusty: This should be ``foldBack`` I think I missed it in my dehaskellize PR. Note that the signature (params order) of Haskell's ``foldr`` is not the same as ``foldBack``.

src/FSharpPlus/Validations.fs
```diff
@@ -0,0 +1,137 @@
...
+
+  let inline traverse f = function 
+    |AccSuccess a -> AccSuccess <!> f a
+    |AccFailure e -> result (AccFailure e)
+
+  let inline bimap f g = function
+    |AccFailure e -> AccFailure (f e)
+    |AccSuccess a -> AccSuccess (g a)
+
+  let inline bifoldr f g x = function
```
gusty: We don't have bifolds, but it's fine to have it as a standalone function and nothing stops us from adding it in the future. But again here, we'll have to change it to something like ``biFoldBack``.
Alternatively we can remove it for the moment.

src/FSharpPlus/Validations.fs
```diff
@@ -0,0 +1,137 @@
...
+
+  let appAccValidation (m:'err -> 'err -> 'err) (e1':AccValidation<'err,'a>) (e2':AccValidation<'err,'a>) =
+    match e1',e2' with
+    |AccFailure e1 , AccFailure e2 -> AccFailure (m e1 e2)
+    |AccFailure _  , AccSuccess a2 -> AccSuccess a2
+    |AccSuccess a1 , AccFailure _  -> AccSuccess a1
+    |AccSuccess a1 , AccSuccess _  -> AccSuccess a1
+
+  let toResult x :Result<_,_> = match x with AccSuccess a -> Ok a | AccFailure e -> Error e
+  let fromResult (x :Result<_,_>) = match x with Ok a -> AccSuccess a | Error e -> AccFailure e
```
gusty: rename to ``ofResult`` ?

src/FSharpPlus/Validations.fs
```diff
@@ -0,0 +1,137 @@
...
+
+/// Validate's the [a] with the given predicate, returning [e] if the predicate does not hold.
+///
+/// This can be thought of as having the less general type:
+///
+/// validate : 'e -> ('a -> bool) -> 'a -> AccValidation<'e, 'a>
+///
+let validate (e:'e) (p:('a -> bool)) (a:'a) : AccValidation<'e,'a> = if p a then AccSuccess a else AccFailure e
+
+/// validationNel : Choice<'a,'e> -> AccValidation (NonEmptyList<'e>) a
```
gusty: Change Choice to Result

src/FSharpPlus/Validations.fs
wallymathieu: Let's call it `Validation`

wallymathieu:
Thanks for the review @gusty !

src/FSharpPlus/Validations.fs
wallymathieu: I've renamed it

src/FSharpPlus/Validations.fs
```diff
@@ -0,0 +1,143 @@
...
+/// that would violate monad rules.
+///
+/// An example of typical usage can be found <https://github.com/qfpl/validation/blob/master/examples/src/Email.hs here>.
+///
+type Validation<'err,'a> =
+  | Failure of 'err
+  | Success of 'a
+
+module Validation=
+  let inline map (f:'T->'U)=function
```
gusty: This is probably not necessary to be inline.

src/FSharpPlus/Validations.fs
```diff
@@ -0,0 +1,143 @@
...
+  ///
+  /// There is nothing wrong with using this function, it just does not make a
+  /// valid Monad instance.
+  let inline bind (f:'T->Validation<_,_>) x :Validation<_,_>=
+      match x with 
+      | Failure e -> Failure e
+      | Success a -> f a
+
+  /// orElse v a returns 'a when v is Failure, and the a in Success a.
+  let inline orElse v (a:'a) =
```
gusty: Same here, and ``bind``.

src/FSharpPlus/Validations.fs
```diff
@@ -0,0 +1,143 @@
...
+  // as Applicative
+  static member Return            x = Success x
+  static member inline (<*>)      (f:Validation<_,'T->'U>, x:Validation<_,'T>) : Validation<_,_> = 
+        Validation.apply f x
+  // as Alternative (inherits from Applicative)
+  static member inline get_Empty () = Failure ( getEmpty() )
+  static member inline (<|>) (x:Validation<_,_>, y:Validation<_,_>) = Validation.appValidation Control.Append.Invoke x y
+
+  // as Functor
+  static member Map        (x : Validation<_,_>, f) = Validation.map f x
```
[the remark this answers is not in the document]
wallymathieu: It should

gusty:
Good job. No doubt this is a good addition for F#+.

wallymathieu:
I'll squash these changes then 😄

PR #105, opened 2018-04-12
Adding ChoiceT

### LINKED ISSUE #99: Possibly create a ChoiceT Monad Transformer
### Description

While I do see a `ResultT` transformer, several .NET libraries still use the (admittedly outdated) `Choice` type (e.g. `Async.Catch`).  The `Choice` type is incompatible with the `Result` type. 

### Repro steps

N/A

### Expected behavior

N/A

### Actual behavior

N/A

### Known workarounds

I suppose I could create a function called `choiceToResult` like this

```
let choiceToResult myThingy =
    match myThingy with
    | Choice1Of2 x -> Ok x
    | Choice2Of2 ex -> Error ex
```
However, this is a bit clunky since I'd have to pipe every `Choice` result into this function.

wallymathieu:
- ChoiceT as same type of implementation as ResultT

This is intended to fix #99

### DIFF
--- a/src/FSharpPlus/Error.fs
+++ b/src/FSharpPlus/Error.fs
@@ -68,3 +68,60 @@ type ResultT<'``monad<'result<'t,'e>>``> with
 
     static member inline get_Get()  = lift get         : '``ResultT<'MonadState<'S,Result<_,'E>>>``
     static member inline Put (x:'S) = x |> put |> lift : '``ResultT<'MonadState<'S,Result<_,'E>>>``
+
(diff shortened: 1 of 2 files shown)


wallymathieu:
@Tombert what do you think?

Tombert:
@wallymathieu I'm hardly an expert on this code, but from what I'm seeing there doesn't seem to be anything objectionable.

wallymathieu:
Thanks! 😃

PR #113, opened 2018-05-19
Prism result

wallymathieu:
- Adding tests to verify prisms
- Adding comments for `option` and `either`

Should be squashed if accepted, since there is a bit of noise on the branch

### DIFF
--- a/src/FSharpPlus/Operators.fs
+++ b/src/FSharpPlus/Operators.fs
@@ -14,7 +14,9 @@ module Operators =
     let inline uncurry f (x, y) = f x y
     let inline (</) x = (|>) x
     let inline (/>) x = flip x
+    /// Case analysis for the Result type. If the value is Ok a, apply the first function to a; if it is Error b, apply the second function to b.
     let inline either f g = function Ok x   -> f x | Error x -> g x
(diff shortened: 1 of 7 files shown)


src/FSharpPlus/Operators.fs
```diff
@@ -14,7 +14,9 @@ module Operators =
     let inline uncurry f (x, y) = f x y
     let inline (</) x = (|>) x
     let inline (/>) x = flip x
+    /// Case analysis for the Result type. If the value is Ok a, apply the first function to a; if it is Error b, apply the second function to b.
```
wallymathieu: I've modified the comment from Haskell core source and modified it.

wallymathieu:
Well, as you can see of the commits, I tried out changing the signature. The effect was that the tests wouldn't compile.

wallymathieu:
If the semantics is that of choice, perhaps we should try out to change it back to choice and see what happens?

gusty:
That's probably the right thing to do.
But since we're now in 1.0.0 we have to be careful, that would be a breaking change.

wallymathieu:
I agree. So the only changes in this PR are in order to verify that the existing behavior still works (and some comments).

PR #115, opened 2018-05-27
Indexable, Foldable and readonly collections

wallymathieu:
Adding some tests in order to demonstrate that readonly collection works with f#+

### DIFF
--- a/tests/FSharpPlus.Tests/General.fs
+++ b/tests/FSharpPlus.Tests/General.fs
@@ -1,6 +1,7 @@
 ﻿namespace FSharpPlus.Tests
 
 open System
+open System.Collections.ObjectModel
 open FSharpPlus
 open FSharpPlus.Data
 open FSharpPlus.Control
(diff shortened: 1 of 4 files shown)


tests/FSharpPlus.Tests/General.fs
```diff
@@ -500,6 +501,33 @@ module Foldable =
...
 
+    [<Test>]
+    let intersperse() =
+        Assert.AreEqual ("a,b,c,d,e", intersperse ',' "abcde")
+        Assert.AreEqual (["a";",";"b";",";"c";",";"d";",";"e"], intersperse "," ["a";"b";"c";"d";"e"])
+
+    [<Test>]
+    let readOnlyIntercalate() =
+        Assert.AreEqual ("Lorem, ipsum, dolor", intercalate ", " ["Lorem"; "ipsum"; "dolor"])
+        Assert.AreEqual ("Lorem, ipsum, dolor", intercalate ", " (ReadOnlyCollection( ["Lorem"; "ipsum"; "dolor"] |> List.toArray )))
```
gusty: Any good reason why instead of:

`["Lorem"; "ipsum"; "dolor"] |> List.toArray`

use array brackets directly `[|` `|]` and save some keystrokes?

`[|"Lorem"; "ipsum"; "dolor"|]`

wallymathieu: No good reason at all. 😅

tests/FSharpPlus.Tests/General.fs
```diff
@@ -580,6 +612,14 @@ module Indexable =
 
         ()
 
+    [<Test>]
+    let tryItemReadonly() =
+        let d = ReadOnlyDictionary( Map.ofSeq [1, "one"; 2, "two"] :> IDictionary<_,_> )
```
gusty: Here you can just do `dict [1, "one"; 2, "two"]` and you'll get the IDictionary

wallymathieu: Thanks! 👍

PR #125, opened 2018-08-19
- Tuple size limitation

gusty:
This removes tuple size limitation.
It also makes available the overloads for both syntactic and system tuples.

The only limitation is for lenses, because MapItemX functions will start returning system tuples for tuples of size greater than 7, when writing.

This is due a wart in the current design of the F# compiler, which only converts tuples when size is less than 8.

### DIFF
--- a/docsrc/content/abstraction-monoid.fsx
+++ b/docsrc/content/abstraction-monoid.fsx
@@ -57,10 +57,8 @@ From .Net/F#
  -  ``Set<'T>``
  -  ``Map<'T,'U>``
  -  ``TimeSpan`` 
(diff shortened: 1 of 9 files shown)


gusty:
Yes, actually it's a workaround for this:

> This is due a wart in the current design of the F# compiler, which only converts tuples when size is less than 8.

It works fine as long as we know in advance the return type, typically when it's the same as the input.

Let's keep this explanation here, for future references.

wallymathieu:
Yes, I think I've seen that function defined somewhere else in a different code base (and asked questions about it) 😉

PR #126, opened 2018-09-04
Adding takeWhile, skipWhile and sortByDescending

wallymathieu:
I've added 
- takeWhile
- skipWhile
- sortByDescending

### DIFF
--- a/src/FSharpPlus/List.fs
+++ b/src/FSharpPlus/List.fs
@@ -27,26 +27,6 @@ module List =
         loopM xs
 
     let inline replicateM count (initial: '``Applicative<'T>``)  = sequence (List.replicate count initial)
-    /// <summary>This function finds the maxima of a list of values, preserving the sequence of retained values from the supplied list.</summary>
-    /// <param name="projection">A function that transforms an element of the list into a comparable key.</param>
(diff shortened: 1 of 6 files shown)


wallymathieu:
I've removed the generic minimaBy and maximaBy.

wallymathieu:
What I can see from the implementation of `List.sortBy` and `Seq.sortBy` they are both implemented by first copying the data to an array and then sorting in place.

src/FSharpPlus/List.fs
```diff
@@ -27,7 +27,26 @@ module List =
         loopM xs
 
     let inline replicateM count (initial: '``Applicative<'T>``)  = sequence (List.replicate count initial)
-
+    /// <summary>This function finds the maxima of a list of values, preserving the sequence of retained values from the supplied list.</summary>
```
gusty: This looks fine, but it should be in Extensions.fs instead.
This is my fault, again because lack of design guidelines, but just move it there. Non-generic functions are mostly there.

wallymathieu: Adding a comment about it here could help. What's the purpose of these modules?

PR #136, opened 2018-10-31
Add polyvariadic memoization

### LINKED ISSUE #88: Memoization
Hi there :slightly_smiling_face: 

[This](https://www.youtube.com/watch?v=fNH0Ql8_Zmo) looks very useful. ^-^

cmeeren:
Fixes #88 

Note that I tried making the helper type `T` private, but found no way to make it compile. We might consider giving it an awful name that is less likely to collide with anything user-defined.

### DIFF
--- a/src/FSharpPlus/Memoization.fs
+++ b/src/FSharpPlus/Memoization.fs
@@ -7,16 +7,20 @@ module Memoization =
     // From https://gist.github.com/gusty/70e5af737f2f6aed2bc0303a2e17c7d7
 
     open System.Collections.Concurrent
-  
+
     // Key wrapper to allow null values, since dict keys can't be null
     [<Struct>]
     type MemoizationKeyWrapper<'a> = MemoizationKeyWrapper of 'a
 
-    type T = T with static member getOrAdd (cd:ConcurrentDictionary<MemoizationKeyWrapper<'a>,'b>) (f:'a -> 'b) k = 
-      cd.GetOrAdd (MemoizationKeyWrapper k, (fun (MemoizationKeyWrapper x) -> x) >> f)
+    type MemoizeConcurrentHelper = MemoizeConcurrentHelper with
+        static member getOrAdd (cd:ConcurrentDictionary<MemoizationKeyWrapper<'a>,'b>) (f:'a -> 'b) k =
+            cd.GetOrAdd (MemoizationKeyWrapper k, (fun (MemoizationKeyWrapper x) -> x) >> f)
 
-    let inline memoize (f:'``(T1 -> T2 -> ... -> Tn)``): '``(T1 -> T2 -> ... -> Tn)`` = (T $ Unchecked.defaultof<'``(T1 -> T2 -> ... -> Tn)``>) f
+    let inline memoize (f:'``(T1 -> T2 -> ... -> Tn)``): '``(T1 -> T2 -> ... -> Tn)`` =
+        (MemoizeConcurrentHelper $ Unchecked.defaultof<'``(T1 -> T2 -> ... -> Tn)``>) f
 
-    type T with     
-        static member        ($) (_:obj,  _: 'a -> 'b) = T.getOrAdd (ConcurrentDictionary ())
-        static member inline ($) (T, _:'t -> 'a -> 'b) = T.getOrAdd (ConcurrentDictionary ()) << (<<) memoize
+    type MemoizeConcurrentHelper with
+        static member ($) (_:obj,  _: 'a -> 'b) =
+            MemoizeConcurrentHelper.getOrAdd (ConcurrentDictionary ())
+        static member inline ($) (MemoizeConcurrentHelper, _:'t -> 'a -> 'b) =
+            MemoizeConcurrentHelper.getOrAdd (ConcurrentDictionary ()) << (<<) memoize
--- a/src/FSharpPlus/Memoization.fs
+++ b/src/FSharpPlus/Memoization.fs
@@ -8,10 +8,15 @@ module Memoization =
 
     open System.Collections.Concurrent
   
-    type T = T with static member getOrAdd (cd:ConcurrentDictionary<_,'t>) (f:_ -> _) k = cd.GetOrAdd (k, f)
-  
+    // Key wrapper to allow null values, since dict keys can't be null
+    [<Struct>]
+    type MemoizationKeyWrapper<'a> = MemoizationKeyWrapper of 'a
+
+    type T = T with static member getOrAdd (cd:ConcurrentDictionary<MemoizationKeyWrapper<'a>,'b>) (f:'a -> 'b) k = 
+      cd.GetOrAdd (MemoizationKeyWrapper k, (fun (MemoizationKeyWrapper x) -> x) >> f)
+
     let inline memoize (f:'``(T1 -> T2 -> ... -> Tn)``): '``(T1 -> T2 -> ... -> Tn)`` = (T $ Unchecked.defaultof<'``(T1 -> T2 -> ... -> Tn)``>) f
-  
+
     type T with     
         static member        ($) (_:obj,  _: 'a -> 'b) = T.getOrAdd (ConcurrentDictionary ())
         static member inline ($) (T, _:'t -> 'a -> 'b) = T.getOrAdd (ConcurrentDictionary ()) << (<<) memoize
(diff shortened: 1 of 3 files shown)


wallymathieu:
You could have the default memoize function in the FSharpPlus namespace. That would reduce the likelihood someone seeing `T`. Another way could be to add a module named internal (but still public) with implementation details that you don't want the consumers of the library to see if they open up the module.

src/FSharpPlus/Memoization.fs
```diff
@@ -0,0 +1,17 @@
...
+[<AutoOpen>]
+module Memoization =
+
+    // From https://gist.github.com/gusty/70e5af737f2f6aed2bc0303a2e17c7d7
+
+    open System.Collections.Concurrent
+  
+    type T = T with static member getOrAdd (cd:ConcurrentDictionary<_,'t>) (f:_ -> _) k = cd.GetOrAdd (k, f)
+  
+    let inline memoize (f:'``(T1 -> T2 -> ... -> Tn)``): '``(T1 -> T2 -> ... -> Tn)`` = (T $ Unchecked.defaultof<'``(T1 -> T2 -> ... -> Tn)``>) f
```
wallymathieu: I think @gusty [preferred](https://github.com/fsprojects/FSharpPlus/issues/88#issuecomment-434485033) to have the default memoize function in the default namespace.

cmeeren: Hence the AutoOpen, since you can't define `let` values directly in namespaces. If I have misunderstood, please clarify where you want it placed.

wallymathieu: Sorry, I didn't notice the `AutoOpen`. Perhaps an implementation module named internal or something in order to hide T ?

cmeeren: Are you then referring to simply renaming `module Memoization` to `module Internal` and still have AutoOpen? Or moving the type to `module Internal` and keeping `memoize` in `module Memoize`?

wallymathieu: Would it work to have something like:
```
    module Internal =
       open System.Collections.Concurrent
  
       type T = T with static member getOrAdd (cd:ConcurrentDictionary<_,'t>) (f:_ -> _) k = cd.GetOrAdd (k, f)
       type T with     
          static member        ($) (_:obj,  _: 'a -> 'b) = T.getOrAdd (ConcurrentDictionary ())
          static member inline ($) (T, _:'t -> 'a -> 'b) = T.getOrAdd (ConcurrentDictionary ()) << (<<) 
    open Internal
    let inline memoize (f:'``(T1 -> T2 -> ... -> Tn)``): '``(T1 -> T2 -> ... -> Tn)`` = (T $ Unchecked.defaultof<'``(T1 -> T2 -> ... -> Tn)``>) f
  
```
?

wallymathieu: Inside the Memoization module?

wallymathieu: Though, that would be somewhat the same as your suggestion to give `T` an awful name

cmeeren: I'll check, I have just assumed that the `T` extensions need to be defined after `memoize` since it was that way in the original sample.

wallymathieu: Perhaps give it the name `MemoizeConcurrentT` or something ? What does @gusty think?

cmeeren: I'd perhaps go for `MemoizeConcurrentHelper`

wallymathieu: And it's a bit of a mouthful so you'd not expect anyone to use it unintentionally. 😄

cmeeren: Well that entirely depends upon what you imagine adding to the module in the future.

One could of course extend it with more memoization, like a thread-unsafe but more efficient version using `Dictionary`. But other than memoization, I have no idea what you'd want to place here.

wallymathieu: So I guess this pull request should be fine to merge then once the name of `T` is changed?

cmeeren: I also have another issue; currently it fails for `null` inputs (since `ConcurrentDictionary` can't have `null` keys). I'll create a single-case struct DU to wrap the key in, that seems to be the normal approach to overcome that particular limitation in other F# memoization implementations I've seen.

gusty: Cool !
I haven't thought about the `null` case.

wallymathieu:
Thanks for the contribution @cmeeren !

cmeeren:
Happy to have helped!

On another note: Intellisense was completely unusable when making the edits to to being so slow, so I had to rely on guesswork and FAKE (which also takes a while) and iterate a lot. Is abysmal Intellisense performance also a problem when using this library via nuget? (I haven't gotten that far yet; I started looking into it just yesterday.) Because that would certainly put a stopper to my using it.

gusty:
Well, it depends on many factors, mainly on which level of "genericity" you are pushing the code. In this case (the memoization func) it would be relative high.

It also depends on which editor you use (please share) and it makes a big difference if your compiler is older than F# 4.1 which has many compiler perfs improvements in this regard.

gusty:
It also worth clarifying that this lib does not slows down the compiler for the sole action of linking it.
It always depends on which generic use you do, and you can always go with 0 generic, most functions have an equivalent non-generic version and there is a misconception that FSharpPlus is only about generic code when it's more about a base library with a consistent design.

gusty:
Well, everything consumes compiler cycles.

In that sense, using `List.map` vs `map` consumes less compiler cycles as `map`  pushes a bit more the Constraint Solver.

This is true for normal F# code, in general, using more generic functions and less type annotations puts more pressure on the type inference.

Now, it all depends on how far the compiler has to go in order to infer those types. For example:

If you do: `let x = 5 + 5` type inference doesn't travel that much as in `let inline x = x + x`, so in this library this could be equivalent to `let x = map string [1]` and `let inline x = map string x`.

My personal advice is to use, generic functions in a judicious way, depending on the different scenarios.

One of the main reasons of the generic functions are the operators (therefore the original naming of the module): there is not a big win in having `map string [1]` vs `List.map [1]`, just saves a few keystrokes, but there is big win in having a generic `>>=` or `<*>` since common operators loose they raison d'etre if they're meant to be used with a module prefix (ie: `List.(<*>)`).

Also I note that the F# compiler is currently very slow on tooling on big projects, even if they don't use this library at all and no generic functions.

gusty:
For sure this library assumes the user will take a good compromise and abuse at its own risk.

On the other hand, I'm not a big fan of driving the user compromise by forbidding or omitting stuff (like functions, combinators) that may lead to abuse and reaching a bad code base.

But I worth noting that I don't think that there's much of a risk of slowing down significantly the compiling by using `let x = map string [1]` instead of `let x = List.map string [1]`, although there is a difference in compiler ticks, but I don't think it makes it prohibitive or even noticeable in cases like that.

Using the `memoize` function with many arguments for sure will start putting some noticeable pressure over the constraint solver, to give you an example on the other extreme.

But, I would be more worried about getting a clean code base, and not abuse of generic stuff un-necessarily, in cases where all it does is compromise readability. This is hard to define, as readability depends mainly on the reader ;)

cmeeren:
> This is hard to define, as readability depends mainly on the reader ;)

Indeed it does. I work at a .NET company where I currently am the only one doing F#, so I need to keep things somewhat readable to the non-FPs in the company.

wallymathieu:
Why I've used f# for mostly scripting and analysis in .net company. If you write libraries then you can hide the f# specifics behind a c# compatible API. As long as the library is small and stable enough, they might not feel the need to modify it.

cmeeren:
I'm creating several microservices entirely in F#, so sooner or later someone else will have to get their hands dirty. :)

ShalokShalom:
^-^

:grin: Thanks a lot

ShalokShalom:
> Also I note that the F# compiler is currently very slow on tooling on big projects

Do we know why?

wallymathieu:
@cmeeren If you provide a nice API surface with nice API docs, they might have less to complain about. :wink: Swagger and GraphQL can help you provide a nice frontend experience. As long as you have a selected subset of f# you could get them to pull request small changes.

PR #141, opened 2018-11-01
Suggestion about docs

wallymathieu:
Have I understood it correctly?

### DIFF
--- a/src/FSharpPlus/Builders.fs
+++ b/src/FSharpPlus/Builders.fs
@@ -83,7 +83,7 @@ module Builders =
             Using.Invoke (p.GetEnumerator () :> IDisposable) (fun enum ->
                 let enum = enum :?> IEnumerator<_>
                 this.While (enum.MoveNext, fun () -> rest enum.Current) : '``MonadPlus<'U>``)
-    
+
     type MonadFxStrictBuilder () =
         inherit StrictBuilder ()
         member inline __.Zero () = result ()                                       : '``Monad<unit>``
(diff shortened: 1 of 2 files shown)


cmeeren:
AFAIK users won't see these comments, since users use the instances `monad`, `monad'`, etc.

src/FSharpPlus/Builders.fs
```diff
@@ -83,7 +87,9 @@ module Builders =
             Using.Invoke (p.GetEnumerator () :> IDisposable) (fun enum ->
                 let enum = enum :?> IEnumerator<_>
                 this.While (enum.MoveNext, fun () -> rest enum.Current) : '``MonadPlus<'U>``)
-    
+
+    /// Strict (non lazy) monadic builder. Use when you to avoid potential side-effects due to additive behavior of monad plus.
```
cmeeren: Typo: "when you **want** to"

wallymathieu:
Indeed, it would be better to have these docs in the github documentation page instead.

wallymathieu:
http://fsprojects.github.io/FSharpPlus/computation-expressions.html

src/FSharpPlus/Builders.fs
gusty: This sentence "Use when you to avoid potential side-effects due to additive behavior of monad plus." is wrong.
Actually it's exactly the opposite, this is when you want to have side effects, instead of additive behavior.

src/FSharpPlus/Builders.fs
```diff
@@ -114,6 +121,9 @@ module Builders =
                 if isReallyDelayed then this.While (enum.MoveNext, Delay.Invoke (fun () -> rest enum.Current))
                 else this.strict.While (enum.MoveNext, fun () -> rest enum.Current))
 
+    /// Lazy monadic builder. Use when you to avoid potential side-effects due to additive behavior of monad plus.
```
gusty: Same here.

wallymathieu:
I'm thinking that perhaps it's better not to merge this pull request since it doesn't really improve the docs.

gusty:
Alternatively you can move those XML docs to the instances `monad` `.fx`, not sure TBH if it makes sense.

Definitely we need more doc work for CEs, with examples and also more tests.

PR #143, opened 2018-11-03
Manipulate Map using lenses

wallymathieu:
Porting lens tutorial to f#

## TODO

- [x] Manipulate `Map`


wallymathieu:
Perhaps what's here is enough for now, what do you think @gusty ?

PR #146, opened 2018-11-11
Reimplemented union on dictionaries and maps for performance improvem…

pnobre:
- Reimplemented union on dictionaries and maps using Linq which carries some performance improvements over the current implementation. unionWith will stay as is, because of the combiner function.
- Implemented intersect and intersectWith (preferred intersect instead of join to follow the naming on F# Set).

### DIFF
--- a/src/FSharpPlus/Extensions.fs
+++ b/src/FSharpPlus/Extensions.fs
@@ -378,14 +378,7 @@ module Map =
 
     /// Returns the union of two maps, preferring values from the first in case of duplicate keys.
     let union (source: Map<'Key, 'T>) (altSource: Map<'Key, 'T>) = 
-        Enumerable
-          .Union(
-            source, 
-            altSource,
-            { new IEqualityComparer<KeyValuePair<'Key,'T>> with 
-                      member __.Equals ((a:KeyValuePair<'Key,'T>),(b:KeyValuePair<'Key,'T>)) : bool = a.Key = b.Key
-                      member __.GetHashCode (a:KeyValuePair<'Key,'T>) = a.Key.GetHashCode () })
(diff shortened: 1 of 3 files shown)


gusty:
Could you submit a separate PR for intersects ?

src/FSharpPlus/Extensions.fs
```diff
@@ -1,6 +1,8 @@
 namespace FSharpPlus
 
 open System
+open System.Collections.Generic
```
gusty: This seems to be the reason why your PR is failing.
By opening this namespace at global level, `List` ( `ResizeArray` in F# jargon) shadows F# `List`.

Try opening it locally.

src/FSharpPlus/Extensions.fs
```diff
@@ -370,19 +372,20 @@ module Map =
...
 
-    /// Returns the union of two dictionaries, using the combiner function for duplicate keys.
+    /// Returns the union of two maps, using the combiner function for duplicate keys.
     let unionWith combiner (source1: Map<'Key, 'Value>) (source2: Map<'Key, 'Value>) =
         Map.fold (fun m k v' -> Map.add k (match Map.tryFind k m with Some v -> combiner v v' | None -> v') m) source1 source2
 
     /// Returns the union of two maps, preferring values from the first in case of duplicate keys.
-    let union (source: Map<'Key, 'T>) (altSource: Map<'Key, 'T>) = unionWith (fun k _ -> k) source altSource
-
+    let union (source: Map<'Key, 'T>) (altSource: Map<'Key, 'T>) =
```
gusty: If this is a reversion, could you leave exactly as it was?

This is mainly to reduce the signal to noise ratio of the changes?

pnobre: done.

PR #150, opened 2019-01-02
Update travis yml and paket restore targets

wallymathieu:
Issue with using pack on travis due to paket. Since the main source of packages is Appveyor, there is little need to generate package on Travis.


wallymathieu:
This is to big of a change, probably better to just change target in travis.

PR #159, opened 2019-07-19
Extend string module - Part 1

### LINKED ISSUE #158: Add more string functions
### Description

Add more string functions in Extensions, to the String module.

I propose we compile first a list of nice-to-have functions.

### Related information 

This is related to https://github.com/fsharp/fslang-suggestions/issues/102 and it's motivated by a discussion on slack with @cannorin and @ellersoft

cannorin:
Related:  fsharp/fslang-suggestions#102 
Closes #158.

This PR adds more string functions to the String module (in Extensions.fs), and use them as the implementation of existing generic functions.

"Part 1" means we have some functions left to be discussed in detail and they are not going to be added in this PR.

See #158 for the discussion.

TODO:

* Add the following functions ((*) Generic version already exists):
  - [x] findIndex / tryFindindex / findSliceIndex / tryFindSliceIndex
  - [x] padLeft / padRight / padLeftWith / padRightWith
  - [x] trim / trimStart / trimEnd : taking an array.
  - [x] toArray / ofArray / toList / ofList / toSeq / ofSeq (*)
  - [x] item/tryItem (*)
  - [x] rev (*)
  - [x] take / skip / takeWhile / skipWhile / truncate / drop (*)
  - [x] toCodePoints / ofCodePoints
  - [x] getBytes : accept always an encoding.

* Replace the implementation of following generic functions with the new ones:
  - [x] toArray / ~~ofArray~~ / toList / ofList / toSeq / ofSeq
  - [x] item/tryItem
  - [x] ~~rev~~ *skipped because adding it alone would break the existing overload resolution.*
  - [x] take / skip / takeWhile / skipWhile / truncate / drop

* Add some tests for: (skipped trivial functions)
  - [x] ~~findIndex / tryFindindex / findSliceIndex / tryFindSliceIndex~~
  - [x] ~~padLeft / padRight / padLeftWith / padRightWith~~
  - [x] ~~trim / trimStart / trimEnd~~
  - [x] ~~toArray / ofArray / toList / ofList / toSeq / ofSeq~~
  - [x] ~~item/tryItem~~
  - [x] ~~rev~~
  - [x] ~~take / skip / takeWhile / skipWhile / truncate / drop~~
  - [x] toCodePoints / ofCodePoints
  - [x] ~~getBytes~~

### DIFF
--- a/src/FSharpPlus/Extensions.fs
+++ b/src/FSharpPlus/Extensions.fs
@@ -348,11 +348,15 @@ module String =
     /// Returns a string that have at most N characters from the beginning of the original string.
     /// It returns the original string if it is shorter than count.
     let truncate count (source: string) =
-        if String.length source <= count then source else take count source
+        if count < 1 then String.Empty
+        else if String.length source <= count then source
+        else take count source
     /// Returns a string that drops first N characters of the original string.
     /// When count exceeds the length of the string it returns an empty string.
     let drop     count (source: string) =
-        if String.length source >= count then String.Empty else skip count source
+        if count < 1 then source
+        else if String.length source >= count then String.Empty
+        else skip count source
 
     let findIndex    (char: char) (source: string) = source.IndexOf char
     let tryFindIndex (char: char) (source: string) =
@@ -365,7 +369,7 @@ module String =
         if index = -1 then None else Some index
 
     /// Converts the string to an array of Int32 code-points (the actual Unicode Code Point number).
-    let toCodePoints (encoding : System.Text.Encoding option) (source : string) : seq<int> =
+    let toCodePoints (source : string) : seq<int> =
         let mapper i c =
             // Ignore the low-surrogate because it's already been converted
             if c |> Char.IsLowSurrogate then None
--- a/src/FSharpPlus/Extensions.fs
+++ b/src/FSharpPlus/Extensions.fs
@@ -345,6 +345,24 @@ module String =
                 i <- i + 1
             if i = 0 then ""
             else source |> skip i
+    /// Returns a string that have at most N characters from the beginning of the original string.
+    /// It returns the original string if it is shorter than count.
+    let truncate count (source: string) =
+        if String.length source <= count then source else take count source
+    /// Returns a string that drops first N characters of the original string.
+    /// When count exceeds the length of the string it returns an empty string.
+    let drop     count (source: string) =
+        if String.length source >= count then String.Empty else skip count source
+
+    let findIndex    (char: char) (source: string) = source.IndexOf char
+    let tryFindIndex (char: char) (source: string) =
+        let index = findIndex char source
+        if index = -1 then None else Some index
+
+    let findSliceIndex    (slice: string) (source: string) = source.IndexOf slice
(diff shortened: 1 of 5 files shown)


cannorin:
hmm the CI script on Travis seems to be broken

cannorin:
This `push -f` is for correctly co-authoring @EBrown8534.

src/FSharpPlus/Extensions.fs
```diff
@@ -295,6 +295,93 @@ module String =
...
+        if count < 1 then source
+        else if String.length source >= count then String.Empty
+        else skip count source
+
+    let findIndex    (char: char) (source: string) = source.IndexOf char
+    let tryFindIndex (char: char) (source: string) =
+        let index = findIndex char source
+        if index = -1 then None else Some index
+
+    let findSliceIndex    (slice: string) (source: string) = source.IndexOf slice
```
gusty: I wonder whether should we throw here, instead of returning `-1`

cannorin: Ah it should throw `ArgumentException` as `List.findIndex` do

cannorin:
As FAKE build script is not working on my machine, I've run NUnit manually on the test assembly and confirmed that the test about `toCodePoints / ofCodePoints` I added passes.

src/FSharpPlus/Extensions.fs
```diff
@@ -295,6 +295,103 @@ module String =
...
+    let padLeft totalLength (source: string) = source.PadLeft totalLength
+    /// Pads the beginning of the given string with a specified character so that it has a specified total length.
+    let padLeftWith totalLength paddingChar (source: string) = source.PadLeft (totalLength, paddingChar)
+    /// Pads the end of the given string with spaces so that it has a specified total length.
+    let padRight totalLength (source: string) = source.PadRight totalLength
+    /// Pads the end of the given string with a specified character so that it has a specified total length.
+    let padRightWith totalLength paddingChar (source: string) = source.PadRight (totalLength, paddingChar)
+
+    /// Removes all leading and trailing occurrences of specified characters from the given string.
+    let trim      (trimChars: char seq) (source: string) = source.Trim (Seq.toArray trimChars)
```
gusty: I'm a bit unsure whether the `trimChars` param should be a `char seq`, a `char array` or a `string`. Any arguments for or against any of them?

cannorin: I think `char seq` is the best choice here as it can receive `char list`, `Set<char>`, etc

gusty: Yes, that's a good argument for `char seq`.

As for `char []` I would say it's a bit more performant, anyway the list is expected to be short.

For `string` it's interesting to note that it matches the collection type. Here I'm more thinking again in generic terms, what if we decide in the future to create a trim for lists, arrays ...

cannorin: When generalized, `trim` can be thought as a variant of `except`. It takes `'a seq` as elements to exclude regardless of collection type. I think it would make more sense to follow the same design.

PR #164, opened 2019-07-27
Update 'Learn You a Haskell' sample

paullucas:
Changes:
- Initial whitespace cleanup
- Additional comments denoting result values
- Type annotation for `readMaybe` (L:494)


gusty:
I like the idea of having the results in the file.

I wonder whether this sample should be migrated to the docs.

Thanks !

PR #169, opened 2019-07-31
Add Task Applicative instance

TOTBWF:
This PR adds an overload of `<*>` for `Task`, as well as some helpful extensions to the `Task` module.


wallymathieu:
Awesome!

PR #171, opened 2019-08-01
Bump FSharp.Formatting version

TOTBWF:
`FSharp.Formatting-3.0.0-beta8` is no longer in nuget, so we should probably bump the version.


gusty:
I'm getting this error when generating docs:

    error FS0192: internal error: convMethodRef: could not bind to method

However it didn't work before this PR, not sure if the error was the same.

This is in Windows.

PR #172, opened 2019-08-01
Align memoizeN code with F#+

gusty:
Tried to integrate it a bit better at both internal code level and end user.

I would love to have the function in Operators (and the type in Control) but it seems it have to be interleaved in order to delay overload resolution.

### DIFF
--- a/src/FSharpPlus/Memoization.fs
+++ b/src/FSharpPlus/Memoization.fs
@@ -1,5 +1,6 @@
 ﻿namespace FSharpPlus
 
+open FSharpPlus.Internals
 
 [<AutoOpen>]
 module Memoization =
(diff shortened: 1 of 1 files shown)


src/FSharpPlus/Memoization.fs
```diff
@@ -1,26 +1,26 @@
...
 
     // From https://gist.github.com/gusty/70e5af737f2f6aed2bc0303a2e17c7d7
-
     open System.Collections.Concurrent
 
     // Key wrapper to allow null values, since dict keys can't be null
     [<Struct>]
     type MemoizationKeyWrapper<'a> = MemoizationKeyWrapper of 'a
 
-    type MemoizeConcurrentHelper = MemoizeConcurrentHelper with
```
wallymathieu: removing helper looks more to be in line with the conventions found in f#+

Issue #170, opened 2019-08-01
CI should produce nupkgs as artifacts


TOTBWF:
Right now, it is rather tricky to depend on HEAD using paket, as you need to invoke `build.sh`. I propose that the CI should produce a `.nupkg`.

As an aside, it appears that the `NuGet` fake target is borked.

TOTBWF:
Digging a bit deeper, it appears that the root cause of the `pack` breakage is https://github.com/fsprojects/Paket/issues/3540, so we may have to upgrade paket

gusty:
Or drop it and use plain nuget instead.

TOTBWF:
Oddly enough, `paket pack` works just fine for me, it's `msbuild` that is screwing up

wallymathieu:
Should be easier now to update paket. Specified versions of paket are found here:
https://github.com/fsprojects/FSharpPlus/blob/master/build.cmd#L2
https://github.com/fsprojects/FSharpPlus/blob/master/build.sh#L29-L31

TOTBWF:
After smashing my head against this for a few hours, I'm starting to think it would just be better to put out a new release. Thoughts?

gusty:
Yes, let's do that !

gusty:
The new release is out: https://www.nuget.org/packages/FSharpPlus/1.1.0-CI00271

PR #173, opened 2019-08-01
Add nupkg generation to CI

TOTBWF:
This PR adds a nupkg to the appveyor artifacts, fixing #170.

### DIFF
--- a/build.fsx
+++ b/build.fsx
@@ -418,6 +418,7 @@ Target.create "Release" (fun _ ->
     |> Async.RunSynchronously
 )
 
+Target.create "BuildDocs" ignore
 Target.create "BuildPackage" ignore
 
 // --------------------------------------------------------------------------------------
@@ -435,9 +436,8 @@ open Fake.Core.TargetOperators
   ==> "RunTests"
   ==> "NuGet"
   ==> "CopyNuGet"
-  ==> "GenerateReferenceDocs"
-  ==> "GenerateDocs"
   ==> "BuildPackage"
+  ==> "BuildDocs"
   ==> "All"
   =?> ("ReleaseDocs",BuildServer.isLocalBuild)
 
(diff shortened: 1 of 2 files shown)


TOTBWF:
I have no idea why this is failing, any thoughts?

wallymathieu:
Could be that the ordering has changed due to changes in how FAKE orders dependencies. Since it tries to reorder based on the information provided. My guess is that the old way of having a long chain of dependencies that assume a certain order is faulty. FAKE might want to run things in parallel.

wallymathieu:
Note the
```
Shortened DependencyGraph for Target BuildPackage:
```

wallymathieu:
Looks like issue with doc generation https://github.com/fsprojects/Paket/issues/1295, I'll need to reproduce it.

TOTBWF:
Seeing as we already build the docs in travis, it might make sense to just skip the step entirely on appveyor for now. As an aside, we should also probably make sure to bump the package version when doing this

build.fsx
```diff
@@ -433,10 +433,10 @@ open Fake.Core.TargetOperators
   ==> "Build"
   ==> "CopyBinaries"
   ==> "RunTests"
-  ==> "GenerateReferenceDocs"
-  ==> "GenerateDocs"
   ==> "NuGet"
   ==> "CopyNuGet"
+  ==> "GenerateReferenceDocs"
+  ==> "GenerateDocs"
```
wallymathieu: Here you say that BuildPackage depends on GenerateDocs. Perhaps better to move GenerateReferenceDocs and GenerateDocs to a separate chain and let it only depend on CopyBinaries, then add a dependency on GenerateDocs from All.

TOTBWF: That seems more reasonable, will do

wallymathieu:
The failure on travis looks unrelated, I'll create issue around the doc generation. Thanks for the contribution @TOTBWF 😄

PR #174, opened 2019-08-03
Add `findSliceIndex` function to other collections

### LINKED ISSUE #167: Add `findSliceIndex` function to other collections
### Description

Related: https://github.com/fsprojects/FSharpPlus/issues/158#issuecomment-513526420

I propose we add `findSliceIndex / tryFindSliceIndex` [which we recently added to `String` extension](https://github.com/fsprojects/FSharpPlus/blob/master/src/FSharpPlus/Extensions.fs#L371) to other collections' extensions.

```fsharp
val findSliceIndex : slice:'Collection<'a> -> source:'Collection<'a> -> int

val tryFindSliceIndex : slice:'Collection<'a> -> source:'Collection<'a> -> int option
```

It will work just like ~~the existing `findIndex : 'a -> 'Collection<'a> -> int` but it searches a "slice" (i.e. subsequence, sub-collection) instead of an element~~ `String.findIndex`.

cannorin:
Closes #167.

This PR adds `findSliceIndex` to collections other than `String`, which already have one (#159).

See #167 for the discussion.

TODO:

- ~~[x] Add a new reference to `System.Memory`~~ Skipped, will be done in another PR
- ~~[x] Upgrade FSharp.Core to v4.5 or v4.6~~ Skipped, will be done in another PR
- Implement `findSliceIndex` and `tryFindSliceIndex` for:
  - [x] `Seq`
  - [x] `List`
  - [x] `Array`
- [x] Implement generic `findIndex / tryFindIndex` function
- [x] Implement generic `findSliceIndex / tryFindSliceIndex` function
- [ ] Add some tests (if needed)


cannorin:
:arrow_up: fixing wrong commit e-mail address.

cannorin:
:arrow_up: use naive implementation instead of utilizing `Span<'T>`.

src/FSharpPlus/Extensions.fs
```diff
@@ -180,6 +180,62 @@ module Seq =
     open System.Collections.Generic
     let toIReadOnlyList (x: seq<_>) = x |> ResizeArray |> ReadOnlyCollection :> IReadOnlyList<_>
 
+    open System.Linq
+    let private findSliceIndexBase (slice: seq<_>) (source: seq<_>) =
```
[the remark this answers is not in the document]
cannorin: Ah I see :)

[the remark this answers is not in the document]
cannorin: I noticed `Internals.fs` comes before `Extensions.fs` in the fsproj and swapping them does not break the build. It is ok to do so?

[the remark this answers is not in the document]
gusty: Yes

cannorin:
For `findIndex / tryFindIndex` we must resolve #175 first

cannorin:
I did something wrong trying to merge #177. fixing

src/FSharpPlus/Operators.fs
```diff
@@ -345,7 +345,60 @@ module Operators =
...
+    /// The function to test the input elements.
+    /// </param>
+    /// <param name="source">The input foldable.</param>
+    /// <returns> 
+    /// The index of the first element that satisfies the predicate.
+    /// </returns>
+    /// <exception cref="System.ArgumentException">
+    /// Thrown if the predicate evaluates to false for all the elements of the source.
+    /// </exception>
+    let inline findIndex (predicate: 'T -> bool) (source: '``Foldable<'T>``) : int = FindIndex.Invoke predicate source
```
gusty: Maybe I'm thinking too much, but what if we decide later to add a `findIndex` (and/or the other functions in this PR) to a multidimensional array?

The return type won't be a single int anymore, it would be a pair of ints.

If we think about GetSlice it considers a slice a variable type, becuse of the multidimensional arrays.

Of course, this might complicate type inference.

Thoughts?

cannorin: multidimensionals are not foldable anymore, so we must invent a new abstraction to do it

gusty: Yes, I forgot to mention that. Although we might consider them as Foldables, but in that case the return type will be always an int.

If we want to have a generic return type we need to move out of Foldable, maybe Collection?

cannorin: Collections seems fine to me. I'll do it

gusty: Another alternative, is to leave this as it is, and eventually create another set of (more) generic functions, as it's already the case with `nth` and `item`.

Then we can use the plural `indices` instead of `index`, and we can consider the `Indexable` abstraction.

cannorin: Plural `indices` sounds more like `('a -> bool) -> Collection<'a> -> int []` and I'm personally against that. 
Does `foldi` works on multidimentionals? If so, it will make more sense to place `findIndex` functions to them.

gusty: Good question. TBH I can't recall, I did the Indexable abstraction (actually I took it from somewhere) long time ago :)

gusty: Looking at the code, it seems to me that `foldi` for multidimensionals could be added. I think I was lazy, normally that's always the case ;)

cannorin: I think we need to add generic `getSlice` (or `slice`) built on top of `Indexable`. (`item` - `findIndex ((=) index)`, `getSlice` - `findSliceIndex`)

gusty: Sounds good in principle !

cannorin: Anyway, if making the functions having generic return type (rather than `int`) doesn't break something, I think it is good to do it for now. We can later move them to `Indexable` without much pain.

cannorin: Or is it better to leave it as is?

gusty: They might break, as `int` drives type inference. That's why I think having a second set of the functions might be desirable, although you can argue that's overkill.

gusty: Btw: I'm also thinking, apart from multidimensionals arrays, `findIndex` could work on Dictionaries. Not the findSlice families, as they require a notion of contiguity.
Sorry, if I'm thinking too much, it's not my intent to complicate this PR with all this. What you did makes totally sense, I'm just thinking how can we leave room for later improvements.

wallymathieu: Perhaps for ordered dictionary it makes sense

cannorin: I like to handle all the case with just a single set of functions since it is less confusing to me. My idea is to make their return type generic for now to prevent future breaking changes in type signature. Doint it seems not breaking the currently supported types (foldable, string, array, etc).

gusty: Let's try to be pragmatic as I want to merge this I see 2 options:

 - We find and use another name for "index" which means and integer index and just rename these functions.
 - We remove the int annotation and move these functions to Indexable.

cannorin: For the first option, the term "indexer" is also used for the syntax accessing dictionaries so the meaning of "index" is not limited to integer. So I think we don't have to invent a new word.

I complete agree with the second option.

gusty: I agree as well, @wallymathieu any thoughts? Ok for the 2nd?

gusty: In that option 2 we would have to remove the Default1 implementation as it is biased to an `int` index coming from a fold operation. Personally I'm fine with it.

wallymathieu: I'm fine with it as well.

cannorin:
Sorry for being late, now I've moved the functions to `Indexable`.

gusty:
@forki  Would you be able to add @cannorin to fsprojects ?

forki:
added @cannorin to fsharpplus-admin

cannorin:
@forki @gusty joined! thanks :smile:

PR #180, opened 2019-08-11
[WIP] Add missing builder methods for Task

gusty:


src/FSharpPlus/Functor.fs
```diff
@@ -315,7 +315,8 @@ type Delay =
     static member inline Delay (_mthd: Default1, x: unit-> ^I                                             , _: Delay   ) = (^I : (static member Delay : _->_) x) : ^I
     static member inline Delay (_mthd: Default1, _: unit-> ^t when  ^t : null and ^t  : struct            , _          ) = ()
 
-    static member        Delay (_mthd: Default2, x: unit-> _                                              , _          ) = Seq.delay x      : seq<'T> 
+    static member        Delay (_mthd: Default2, x: unit-> _                                              , _          ) = Seq.delay x      : seq<'T>
+    static member        Delay (_mthd: Delay   , x: unit-> Task<_>                                        , _          ) = new Task<_> (fun () -> x().Result) : Task<'T>
```
wallymathieu: I'm not sure that this will delay the task

wallymathieu: Since you already have a potentially running task, the only way here would be to just pass it along?

gusty:
Since Task is not a lazy monad.
Is a bit of a mix, because although it is a computation, the computation starts immediately and most importantly, although there is a way to delay it, users assume it is not delayed.

PR #185, opened 2019-08-18
Free Monads

gusty:
This PR will introduce generic Free Monads !

We need to discuss the names (Pure, Roll, Free, Join), create tests and add sample code.
And of course some docs would be ideal but these last years there were many posts about Free Monads in the context of dependency injection.

I tried coding [this sample](http://www.fssnip.net/7SX/title/Freemonad-interpreter) and works up to the definition of `exF2`,  I still haven't found a workaround but realized that if we use `Map.InvokeOnInstance` and restrict the Free Monads to the user defined types (excluding primitive functor types) it works fully. So that's a B plan.


gusty:
I'm willing to merge this.
I think the only open question at this point is the naming we're using.
I just picked, without thinking too much: `Free` for the type, `Pure` and `Roll` for the DU.

Any thoughts?

cannorin:
I think that naming is fine.

wallymathieu:
I haven't grokked the free monad things, so I can't offer an opinion on the names.

PR #187, opened 2019-08-21
Fix `foldlOf`, and add `maximumOf` and `minimumOf` to Lens

TOTBWF:


gusty:
I don't think it will help further, it was already simplified in `rmap'`.
But it reads better.

TOTBWF:
The fix was changing that `lmap'` to `<<`, as it was screwing up the types

PR #198, opened 2019-09-17
Enabling F#+ code to be tested in F# and Fable

jbotto94:
This PR attempts provide a method for running tests written with F#+ code in both Fable and the F# compiler. The approach used is similar to the one taken by the Fable compiler, which uses Expecto to run tests in F# and Mocha to run tests in JS.

The added code mainly resides in: `tests/FSharpPlusFable.Tests/`

- **tests/FSharpPlusFable.Tests/Util.fs**: Defines the functions for defining basic tests and equalities between different structures. Depending on the compiler used, these are defined using functions from Expecto or Mocha.

- **build.fsx/build.sh**: A simple bash script that runs the tests in both F# and Fable.
(The functions for commands to be run from an fsi file has been partly adapted from: [https://github.com/jet/XRay/blob/master/buildDocs.fsx](url)). The commands are as follows:
  - **"npminstall"** installs the npm dependencies required to run the tests in Fable
  - **"fabletest"** runs the tests in Fable
  - **"fsharptest"** runs the tests in the F# compiler

- **FSharpTests/Extensions.fs**: Defines some sample tests using code from the F#+ Extensions file.

- **Tests.fs**: The main program that runs the tests in F# or Fable.

- **ObjCompare.js**: Defines an equality function in JS for Arrays/Iterables in JS. This is required because Mocha's `Assert.AreEqual` does not properly compare the values of `Sequences/Arrays` that consist of further nested arrays/sequences.


MangelMaxime:
I think should be `objCompare` replaced with `deepEqual` from Mocha.

jbotto94:
@MangelMaxime 
Even though `deepEqual` does work in these 5 simple test cases, I have found it sometimes does not work on some more complex structures that are formed when using Fable to compile some F#+ code.

I have been implementing F#+ functions to be Fable compatible for a future PR and I've found that `deepEqual` fails on more complex objects, including the following:
- `1,1,2,2,3,a2 => f(a1, a2),4,4,5,6` and `1,1,2,2,3,a2 => f(a1, a2),4,4,5,6`
- `[1,1; 2,2; 3,a2 => f(a1, a2); 4,4; 5,6]` and `[1,1; 2,2; 3,a2 => f(a1, a2); 4,4; 5,6]`

I've tried flattening the structures into an `array` type before running the tests. This helps run more tests, but not all.

Therefore, I thought that by using `deepEqual` now would mean possible problems in testing more complex structures in future added test cases.

gusty:
Would another option be to find a repro and report it to Mocha?
Or submit a PR there with the code you did for `objCompare`?

Zaid-Ajaj:
@jbotto94 There is already a package for unit-testing Fable/.NET code, see [Fable.Mocha](https://github.com/Zaid-Ajaj/Fable.Mocha). Don't reinvent the wheel with every library having it's own testing extensions and Fable.Mocha can run the tests both in node.js and in the browser to check if the code is really cross platform 

Second, use F# structural equality to test complex objects and don't rely on JS functions such as `objCompare` and `deepEqual` because they depend on the underlying implementation details of JS representation of the object which might change in different versions of Fable

jbotto94:
@Zaid-Ajaj It would be very easy to transition my tests to use Fable.Mocha since this framework makes use of both Expecto and Mocha and is in the style of the Fable tests. I will have a deeper look into it.

Regarding your second point, I've been experimenting with F# structural equality and `deepEqual`. 
F# structural equality works ever so slightly better than `deepEqual`.
`deepEqual` does not work for the associative case in the test file i provide.
Whilst F# structural equality does work for the cases I have provided, it does not work with some more complex tests involving several data structures. I will give a sample example of some.
I have not yet found an alternative to my function which works for all cases.

PR #160, opened 2019-07-19
Add Fable support

jannesiera:
First step to enable consumption of FSharpPlus from Fable programs is to enable Fable to pick up the nuget.

PR #199, opened 2019-09-19
Surrounding F#+ files with compiler directives

jbotto94:
Surrounding all of the F#+ with compiler directives in a similar way as was done with `Externals.fs` in #197 and #160.


gusty:
I think that in order to add consistency to this PR, code that uses those excluded functions should also be surrounded, otherwise this PR would break Fable compilation.

wallymathieu:
Should the js files be included?

jbotto94:
> Should the js files be included?

They shouldn't. Those are local build files. I have fixed this problem.
This PR should now only alter the core F#+ files, .gitignore and package.json inside FSharpPlusFable.Tests.

src/FSharpPlus/Converter.fs
```diff
@@ -106,28 +123,35 @@ type TryParse with
...
 #if NET35
 #else
+#if !FABLE_COMPILER
     static member        Parse (_: 'T when 'T : enum<_>, _: Parse   ) = fun x ->
         (match Enum.TryParse (x) with
             | (true, v) -> v
             | _         -> invalidArg "value" ("Requested value '" + x + "' was not found.")
         ) : 'enum
 #endif
+#endif
```
wallymathieu: double #endif?

src/FSharpPlus/Functor.fs
```diff
@@ -23,7 +23,9 @@ type Bind =
     static member        (>>=) (source: seq<'T>    , f: 'T -> seq<'U>     ) = Seq.bind f source                                             : seq<'U> 
 #if NET35
 #else
+#if !FABLE_COMPILER
     static member        (>>=) (source: Task<'T>   , f: 'T -> Task<'U>    ) = source.ContinueWith(fun (x: Task<_>) -> f x.Result).Unwrap () : Task<'U>
+#endif
```
wallymathieu: double #endif?

.gitignore
```diff
@@ -202,3 +202,8 @@ docsrc/tools/FSharp.Formatting.svclog
 # Node install directory
 
 tests/FSharpPlusFable.Tests/node_modules/
+
+
+# Fable compilation files
+
+tests/FSharpPlusFable.Tests/FSharpPlusFable
```
wallymathieu: shouldn't this be more specific to exclude only js files? You could add a .gitignore to that folder to exclude the compilation output.

src/FSharpPlus/Converter.fs
jbotto94: If you see there are two compiler directives:
`#if NET35 ... #else ... #endif` and `#if !FABLE_COMPILER ... #endif`
The double `#endif` is for each `#if`.

tests/FSharpPlusFable.Tests/package.json
```diff
@@ -1,6 +1,7 @@
 {
   "private": true,
   "scripts": {
+    "buildFSharpPlus": "fable-splitter ../../src/FSharpPlus -o FSharpPlusFable --commonjs --allFiles",
```
jbotto94: @wallymathieu This is the local target I've added to the local build file of the F#+ Fable tests.

.gitignore
jbotto94: The folder FSharpPlusFable is created when you compile F#+ with Fable through the target I've added. I've tagged you at this line. This makes git ignore any result of compiling F#+ locally.

wallymathieu:
What do you think @gusty, should we merge this one?

gusty:
Ideally we would have another reviewer from the web side.
Since nothing has happened since, I will approve it and we can merge it.
If someone finds a problem with this, we would have to fix it.
CC: @jannesiera

PR #202, opened 2019-10-02
Removing Net35 directives

jbotto94:


gusty:
For the record:

We don't have any evidence of .Net 3.5 use of this library nor have we any target in our CI that makes sure this is properly building, so we'll remove the code for .Net 3.5 for now.

PR #205, opened 2019-12-07
Rename IReadOnlyList.add to trySetItem

wallymathieu:
Also added some documentation while at it

### DIFF
--- a/src/FSharpPlus/Extensions.fs
+++ b/src/FSharpPlus/Extensions.fs
@@ -552,7 +552,7 @@ module IReadOnlyList =
     #if !FABLE_COMPILER
     
     /// Returns a new IReadOnlyList from a given IReadOnlyList, with replaced binding for index.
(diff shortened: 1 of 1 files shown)


gusty:
I'm still not sure which one aligns better: `trySetItem` or `trySetValue`.

I'm gathering evidence of actual naming to see which one is the de-facto standard.

wallymathieu:
Good question, there are some merits to both. Though tryItem is already declared on the module, why perhaps we should rename both in that case

cannorin:
The intuition behind the naming of `tryItem` would be  `get_Item` and `set_Item`, so I think `try{Get,Set}Item` is the best option. But since there are so many `item/tryItem` already in the library, it might not worth renaming all of them to `getItem/tryGetItem`.

wallymathieu:
Yes, tryGetItem would be neat to have.

PR #192, opened 2019-09-02
Type-level literals and Nat-dependent vector/matrix

cannorin:
Closes #178.
Closes #201.

This PR attempts to simulate nat-dependent vector and matrix types through type-level literals.

Things to be discussed:
- Operator naming (currently adding a suffix `^` to avoid problems with globally-defined operators such as `=` and `+`)
- Where to place these things (currently in a separated assembly for faster debugging)
- To be added?

TODO:

- [x] Implement type-level booleans and naturals
- [x] Add type provider for type-level naturals
- [x] Implement nat-dependent vector types
- [x] Implement nat-dependent matrix types as well
- [x] Add some "tests" (which is run by compiling it) if needed

PR #207, opened 2019-12-07
Use strict types for applicatives

gusty:
This will switch to use the internal technique of ~~tupling~~ wrapping in order to avoid flexible types in overloads for the Applicative dispatcher.

Right now there are 2 issues:

 - Types implementing `IEnumerable<_>` would default to the `seq<_>` instance. This might be arguably desired.

 - Types implementing `IEnumerable<_>` would default to the `seq<_>` instance, even if they implement their own applicative instance. This is not acceptable.

### DIFF
--- a/src/FSharpPlus/Functor.fs
+++ b/src/FSharpPlus/Functor.fs
@@ -157,6 +157,7 @@ type Apply =
     static member        ``<*>`` ((f: Result<_,'E>     , x: Result<'T,'E>        , _output: Result<'b,'E>)        , [<Optional>]_mthd: Apply) = Result.apply f x : Result<'U,'E>
     static member        ``<*>`` ((f: Choice<_,'E>     , x: Choice<'T,'E>        , _output: Choice<'b,'E>)        , [<Optional>]_mthd: Apply) = Choice.apply f x : Choice<'U,'E>
     static member inline ``<*>`` ((KeyValue(a: 'Key, f), KeyValue(b: 'Key, x: 'T), _output: KeyValuePair<'Key,'U>), [<Optional>]_mthd: Apply) : KeyValuePair<'Key,'U> = KeyValuePair (Plus.Invoke a b, f x)
+    static member inline ``<*>`` ((struct (a: 'Monoid, f), struct (b: 'Monoid, x: 'T), _output: struct ('Monoid * 'U)), [<Optional>]_mthd: Apply) = struct (FSharpPlus.Control.Plus.Invoke a b, f x)
 
     static member        ``<*>`` ((f: Map<'Key,_>      , x: Map<'Key,'T>         , _output: Map<'Key,'U>)         , [<Optional>]_mthd: Apply) : Map<'Key,'U> = Map (seq {
        for KeyValue(k, vf) in f do
--- a/src/FSharpPlus/Functor.fs
+++ b/src/FSharpPlus/Functor.fs
@@ -140,31 +140,31 @@ type Return =
 type Apply =
     inherit Default1
 
-    static member        ``<*>`` (f: Lazy<'T->'U>     , _: ref<Lazy<'T->'U>>  , x: Lazy<'T>       , [<Optional>]_output: Lazy<'U>        , [<Optional>]_mthd: Apply) = Lazy<_>.Create (fun () -> f.Value x.Value)   : Lazy<'U>
-    static member        ``<*>`` (f: seq<_>           , _: ref<seq<_>>        , x: seq<'T>        , [<Optional>]_output: seq<'U>         , [<Optional>]_mthd: Apply) = Seq.apply  f x                               : seq<'U>
+    static member        ``<*>`` (f: Lazy<'T->'U>     , _: ref<Lazy<'T->'U>>  , x: Lazy<'T>       , [<Optional>]_output: ref<Lazy<'U>>        , [<Optional>]_mthd: Apply) = Lazy<_>.Create (fun () -> f.Value x.Value)   : Lazy<'U>
+    static member        ``<*>`` (f: seq<_>           , _: ref<seq<_>>        , x: seq<'T>        , [<Optional>]_output: ref<seq<'U>>         , [<Optional>]_mthd: Apply) = Seq.apply  f x                               : seq<'U>
     #if !FABLE_COMPILER
-    static member        ``<*>`` (f: IEnumerator<_>   , _: ref<IEnumerator<_>>, x: IEnumerator<'T>, [<Optional>]_output: IEnumerator<'U> , [<Optional>]_mthd: Apply) = Enumerator.map2 id f x : IEnumerator<'U>
+    static member        ``<*>`` (f: IEnumerator<_>   , _: ref<IEnumerator<_>>, x: IEnumerator<'T>, [<Optional>]_output: ref<IEnumerator<'U>> , [<Optional>]_mthd: Apply) = Enumerator.map2 id f x : IEnumerator<'U>
     #endif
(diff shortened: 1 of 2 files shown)


cannorin:
I tested with https://github.com/fsprojects/FSharpPlus/pull/207/commits/fea4b08e8ddb8ddce36da264a6b53d07b8fe4d8d but it seems to failing to redirect.

```fsharp
Microsoft (R) F# Interactive version 10.2.3 for F# 4.5
Copyright (c) Microsoft Corporation. All Rights Reserved.

For help type #help;;

> open FSharpPlus.TypeLits;;
> let v = vector ((fun i -> i + 1), (fun i -> i * 2));;
[<Struct>]
val v : Vector<(int -> int),S<S<Z>>> = [|<fun:v@2>; <fun:v@2-1>|]

> let u = vector (2, 3);;
[<Struct>]
val u : Vector<int,S<S<Z>>> = [|2; 3|]

> v <*> u;;
[<Struct>]
val it : Vector<int,S<S<Z>>> = [|3; 6|]

> open FSharpPlus;;
> v <*> u;;
val it : seq<int> = seq [3; 4; 4; 6]
```

src/FSharpPlus/Functor.fs
```diff
@@ -139,35 +139,32 @@ type Return =
...
+    static member        ``<*>`` (f: Result<_,'E>     , _: ref<Result<_,'E>>         , x: Result<'T,'E>        , [<Optional>]_output: Result<'b,'E>        , [<Optional>]_mthd: Apply) = Result.apply f x : Result<'U,'E>
+    static member        ``<*>`` (f: Choice<_,'E>     , _: ref<Choice<_,'E>>         , x: Choice<'T,'E>        , [<Optional>]_output: Choice<'b,'E>        , [<Optional>]_mthd: Apply) = Choice.apply f x : Choice<'U,'E>
+    static member inline ``<*>`` (KeyValue(a: 'Key, f), _: ref<KeyValuePair<'Key, _>>, KeyValue(b: 'Key, x: 'T), [<Optional>]_output: KeyValuePair<'Key,'U>, [<Optional>]_mthd: Apply) : KeyValuePair<'Key,'U> = KeyValuePair (Plus.Invoke a b, f x)
 
-    static member        ``<*>`` (f: Map<'Key,_>      , x: Map<'Key,'T>         , [<Optional>]_output: Map<'Key,'U>         , [<Optional>]_mthd: Apply) : Map<'Key,'U> = Map (seq {
+    static member        ``<*>`` (f: Map<'Key,_>      , _: ref<Map<'Key,_>>, x: Map<'Key,'T>         , [<Optional>]_output: Map<'Key,'U>         , [<Optional>]_mthd: Apply) : Map<'Key,'U> = Map (seq {
```
wallymathieu: What does wrapping ref<T> help with?

src/FSharpPlus/Functor.fs
```diff
@@ -176,19 +173,24 @@ type Apply =
...
+            ((^M or ^I1 or ^I2 or ^R) : (static member ``<*>`` : _*_*_*_*_ -> _) input1, ref input1, input2, output, mthd)
         call(Unchecked.defaultof<Apply>, f, x, Unchecked.defaultof<'``Applicative<'U>``>)
 
     static member inline InvokeOnInstance (f: '``Applicative<'T->'U>``) (x: '``Applicative<'T>``) : '``Applicative<'U>`` =
         ((^``Applicative<'T->'U>`` or ^``Applicative<'T>`` or ^``Applicative<'U>``) : (static member (<*>) : _*_ -> _) (f, x))
 
+type Apply with
+    static member inline ``<*>`` (f: '``Monad<'T->'U>``      , _: ref<'``Monad<'T->'U>``>      , x: '``Monad<'T>``      , _output: '``Monad<'U>``      , _mthd:Default2) : '``Monad<'U>`` = Bind.InvokeOnInstance f (fun (x1: 'T->'U) -> Bind.InvokeOnInstance x (fun x2 -> Return.InvokeOnInstance (x1 x2)))
+    static member inline ``<*>`` (f: '``Applicative<'T->'U>``, _: ref<'``Applicative<'T->'U>``>, x: '``Applicative<'T>``, _output: '``Applicative<'U>``, _mthd:Default1) : '``Applicative<'U>`` = Apply.InvokeOnInstance f x
+    static member inline ``<*>`` (_: ^t when ^t : null and ^t: struct, _, _: ^u when ^u : null and ^u: struct, _output: ^r when ^r : null and ^r: struct, _mthd: Default1) = id
```
wallymathieu: Or is this the main point of the PR?

src/FSharpPlus/Functor.fs
cannorin: yes that's the point of this PR, F# should not have covariance so the custom type with `IE<_>` implemented should not trigger the overload expecting `ref<_ seq>`

wallymathieu: Aa, thanks @cannorin !

gusty: Just a quick update. At the moment it succeed with `f <*> x` but not with `f <*> x <*> y`, in the latter case still tried to infer `seq`, but at least now it fails.

gusty:
Another update.
The other techinique that can be used in place or combined with wrapping is to add a higher DefaultX marker to the interface overload, the problem here if I do that is that type inference gets worst when resolving to those overloads and needs type annotation. This is currently a noticeable problem with the monad overload.
I'm coming to the conclusion that if there's no way to solve that type inference problem, a good compromise would be to let interfaces type inference degrade, that problem is preferred to not having external instances being resolved, but it would clearly be a breaking change.

gusty:
This is passing now, but there are some breaking specially in the idiom brackets section.
Also the Apply instance for `KeyValuePair` is removed.

Now, if we put in the balance this regression against not having externally defined applicatives implementing interfaces working decently (namely something along a plain `f <*> x <*> y`) it should be clear that this is priority.

How we rollout this change? Possibly we would have to skip a 1.1 version and move directly to a 2.0

The open question is whether this could be improved in order minimize the regression. I spent a crazy amount of time looking at this, if I add the time accumulated in fighting applicatives it's probably the time required to fix the F# compiler to make them work properly for once.

If I have more time, I will keep trying, but otherwise this would have to be merged.

Thoughts?

cannorin:
It's working now!

```fsharp
Microsoft (R) F# Interactive version 10.2.3 for F# 4.5
Copyright (c) Microsoft Corporation. All Rights Reserved.

For help type #help;;

> open FSharpPlus.TypeLits;;
> let v = vector ((fun i -> i + 1), (fun i -> i * 2));;
[<Struct>]
val v : Vector<(int -> int),S<S<Z>>> = [|<fun:v@2>; <fun:v@2-1>|]

> let u = vector (2, 3);;
[<Struct>]
val u : Vector<int,S<S<Z>>> = [|2; 3|]

> v <*> u;;
[<Struct>]
val it : Vector<int,S<S<Z>>> = [|3; 6|]

>  open FSharpPlus;;
> v <*> u;;
[<Struct>]
val it : Vector<int,S<S<Z>>> = [|3; 6|]
```

I suppose very few people seriously uses idiom brackets, so these breaking changes are acceptable.
Since we are introducing big changes such as #185 and #192 and other breaking change #205 I think it is better to move to 2.0 skipping 1.1.

gusty:
Thanks for your feedback. #185 and #192 are not minor things, but at the same time they don't break anything.

Regarding #205 it's not breaking 1.0 since it didn't exists at that time.

So far this is the first serious reason to break 1.0, however I can think of other upcoming breaking changes:

- A new `ListT` monad transformer I have on drafts, that won't be compatible with the existing one.

- A redesign of the computation expression without the "autosense" functionality which users of this library have demonstrated it works only on certain scenarios and create more confusion than other thing.

gusty:
Continued in #569

PR #209, opened 2019-12-14
Use same signature for internal and externals

gusty:


gusty:
Closed in favor of #207 which has less breaking changes.

PR #218, opened 2019-12-20
include style.css and tips.js from F# formatting

wallymathieu:
This is due to the removal of F# formatting to process the docs


wallymathieu:
Not needed

Issue #213, opened 2019-12-18
Make 'either' generic over all bifunctors?


cmeeren:
I have no strong background in CT; sorry if this makes no sense.

I notice that `either` is restricted to the `Result` type. Wouldn't it make more sense to make it generic over all bifunctors? E.g. `Choice<_,_>`.

I don't actually need this, I just noticed it and felt like asking.

gusty:
Yes, they all should be made `Bifoldable`, then `bifold` would be your function ;)

See https://hackage.haskell.org/package/bifunctors-5.4.1/docs/Data-Bifoldable.html

cmeeren:
Makes sense. :) Should `either` be deprecated, then?

gusty:
Yes, deprecated and/or moved as an extension for `Result`.

PR #220, opened 2019-12-28
Bifoldable abstraction

smoothdeveloper:
This is going to be a long and non-optimal journey towards #213.

Opening here to leverage the CI and getting input from people aware of idioms of implementation in this library.

### DIFF
--- a/FSharpPlus.sln
+++ b/FSharpPlus.sln
@@ -1,6 +1,6 @@
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio Version 16
-VisualStudioVersion = 16.0.29613.14
+# Visual Studio 15
+VisualStudioVersion = 15.0.26228.4
 MinimumVisualStudioVersion = 10.0.40219.1
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = ".paket", ".paket", "{63297B98-5CED-492C-A5B7-A5B4F73CF142}"
 	ProjectSection(SolutionItems) = preProject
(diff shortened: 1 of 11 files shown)


gusty:
Looks like there is a Monoid constraint in all Bifoldable functions.

Now, for Result / Choice that constraint is not really an issue, since there's always a single value, so you don't need to fold them into one.

The question is, should we remove that restriction? Or should we keep it?

If we keep it, we'll end up with `bifoldMap` which would be like a generic `either` but with the Monoid restriction.

The monoid restriction is not that bad in F#+ since many types are monoids, including numeric types which in other languages like Haskell are not.

Thoughts ?

smoothdeveloper:
@gusty, to add to your comment, this is what I get in ghci:

> :k Bifoldable
```haskell
Bifoldable :: (* -> * -> *) -> Constraint
```

> :t bifoldMap
```haskell
bifoldMap :: (Bifoldable p, Monoid m) => (a -> m) -> (b -> m) -> p a b -> m
```

> :t bifoldr
```haskell
bifoldr :: Bifoldable p => (a -> c -> c) -> (b -> c -> c) -> c -> p a b -> c
```
> :t bifold
```haskell
bifold :: (Bifoldable p, Monoid m) => p m m -> m
```

The type class itself doesn't have constraint for `Monoid`, but it appears on `bifoldMap` and `bifold`, while `bifoldr` is unrestricted.

gusty:
Regarding the names.
Normally

In Haskell `fold` is called `foldr`
.. `foldback` is `foldl`
.. `sum` is `fold`

So I think we rename `bifold` to `bisum`.

Then `bifold` becames available again. Should we add the Haskell's `foldl` implementation?

wallymathieu:
Perhaps something to add to https://github.com/fsprojects/FSharp.Compatibility once this is merged?

gusty:
I think we can update the reference, specially once we're in v1.1 (very soon I hope)

wallymathieu:
The above looks amenable for ploeh [recent example of [impure/pure/impure](https://blog.ploeh.dk/2019/12/02/refactoring-registration-flow-to-functional-architecture/) since he uses a map that goes over both cases.

gusty:
@wallymathieu I think he speaks there more about Bitraversable, which should be the next think to add ;)

smoothdeveloper:
@gusty just standing on the shoulders of the SRTP giants! 🙂

wallymathieu:
Thanks @gusty ! 😄

PR #223, opened 2020-01-02
Add protect function

gusty:
Adds a function to create a safe version of potentially unsafe functions.

### DIFF
--- a/src/FSharpPlus/Extensions.fs
+++ b/src/FSharpPlus/Extensions.fs
@@ -48,7 +48,7 @@ module Option =
     /// <returns>The resulting option value.</returns>
     let ofResult source = match source with Ok x -> Some x | Error _ -> None
 
-    /// Creates a safe version of the suppiled function, which returns an option<'t> instead of throwing exceptions.
+    /// Creates a safe version of the supplied function, which returns an option<'t> instead of throwing exceptions.
     let protect f x =
         try
             Some (f x)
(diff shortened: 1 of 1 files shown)


src/FSharpPlus/Extensions.fs
```diff
@@ -48,6 +48,12 @@ module Option =
     /// <returns>The resulting option value.</returns>
     let ofResult source = match source with Ok x -> Some x | Error _ -> None
 
+    /// Creates a safe version of the suppiled function, which returns an option<'t> instead of throwing exceptions.
```
jltrem: spelling: suppiled should be supplied

gusty: good catch, thanks !

src/FSharpPlus/Extensions.fs
```diff
@@ -59,6 +65,12 @@ module Result =
     let inline catch f           = function Ok v      -> Ok v     | Error e                 -> (f: 't->_) e : Result<'v,'e>
     let inline either f g        = function Ok v      -> f v      | Error e                 -> g e
 
+    /// Creates a safe version of the supplied function, which returns a Result<'t,exn> instead of throwing exceptions.
+    let protect f x =
```
wallymathieu: I think this makes sense in some cases.

adz:
Thanks for this. Was looking for such a function last week!

Issue #229, opened 2020-01-07
Add result extract


adz:
This extension returns `n` as the default if none: `option f n`
https://github.com/fsprojects/FSharpPlus/blob/master/src/FSharpPlus/Operators.fs#L39

This seems like it's equivalent to (map f >> Option.defaultValue n).

Can we add something similar for Result?

I'd like Result.defaultValue and Result.defaultWith for similarity to existing Option functions but could also be like Haskell fromLeft/fromRight... here as fromOk/fromError
https://hackage.haskell.org/package/base-4.10.0.0/docs/Data-Either.html#v:fromLeft

WDYT?

gusty:
I think we can add the `extract` function, throwing a generic exception message.

We can also add a specific (maybe as an extension method) version of extract that throws )or re-throws) the Error part which is already an exception.

Then for the Haskell's `fromRight`, `fromLeft` we have to agree the names. I propose `getOkWithDefault`, `getErrorWithDefault` and `getChoiceXOf2WithDefault`.

More thoughts?

wallymathieu:
Yes `getOkWithDefault`, `getErrorWithDefault` both for name conventions so would be fine.

adz:
I'm very :+1: on the `extract`. It's what motivated the issue.

However, on the other functions:

There is also `Option.defaultWith` which allows you to run a function rather than a value. In Option case it takes unit, since it is converting from a None.

... but that would make it `getOkWithDefaultWith` which reads poorly.

I think it would be more aligned to say `Result.getOkDefaultValue` and `getOkDefaultWith` along with Error versions.

The other view point is practically, when using the Result type you rarely want to 'throw away' an Ok value keeping the error case, and while it might seem more symmetrical to have both, I think having just `Result.defaultValue` and `Result.defaultWith` would be most discoverable and expected to retain the "Ok". I'd suggest aliasing these functions?

So, my suggestion:
1) `getOkDefaultValue` (aliased as `defaultValue`), `getOkDefaultWith` (aliased as `defaultWith`)
2) `getErrorDefaultValue`, `getErrorDefaultWith`

`With` indicates it takes a mapping function.

gusty:
> The other view point is practically, when using the Result type you rarely want to 'throw away' an Ok value keeping the error case

Do you think the same applies to `Choice` ?

wallymathieu:
One of the nice things about Result type is that you might not want to throw away an Error value either, since otherwise an Option would fit better for model the result.

[...]

PR #232, opened 2020-01-12
+ Some Choice and Result functions

gusty:
This attempts to address some functions requested in #229

### DIFF
--- a/src/FSharpPlus/Extensions.fs
+++ b/src/FSharpPlus/Extensions.fs
@@ -71,8 +71,9 @@ module Result =
             Ok (f x)
         with e -> Error e
 
-    let getOk (source: Result<'T,'Error>) = match source with Ok x -> x | _ -> invalidArg "source" "Result value was Error"
-    let defaultValue value   source = match source with Ok v -> v | _ -> value
+    /// Gets the 'Ok' value. If it's an 'Error' this function will throw an exception.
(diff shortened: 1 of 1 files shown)


src/FSharpPlus/Extensions.fs
```diff
@@ -71,6 +71,16 @@ module Result =
             Ok (f x)
         with e -> Error e
 
+    let getOk (source: Result<'T,'Error>) = match source with Ok x -> x | _ -> invalidArg "source" "Result value was Error"
+    let defaultValue value   source = match source with Ok v -> v | _ -> value
+    let defaultWith defThunk source = match source with Ok v -> v | _ -> defThunk ()
```
adz: I'd have thought we should use an error mapping function rather than a thunk?
WDYT?
```suggestion
    let defaultWith f source = match source with Ok v -> v | e -> f e
```

src/FSharpPlus/Extensions.fs
```diff
@@ -71,6 +71,16 @@ module Result =
             Ok (f x)
         with e -> Error e
 
+    let getOk (source: Result<'T,'Error>) = match source with Ok x -> x | _ -> invalidArg "source" "Result value was Error"
```
adz: I'd go with `get` ... it should probably be discouraged anyway...

gusty: Ok, let's change it to `get`. Yes, it is as unsafe as `get` from option.

PR #236, opened 2020-01-15
Unify Free and FreeNode in a single DU

gusty:
Looks like now that we were forced to use the `Free.run` function there's no big reason for `FreeNode` to mediate.


wallymathieu:
👍

gusty:
Now the only question is, should we make it type, or leave it as a simple DU?

The advantage of a type is that it hides its representation, having the cases visible could be confusing for the end-user.

But, since the constructors are shadowed and also the de-constructors, it shouldn't pose any problems.

Issue #237, opened 2020-01-15
Make 'or' the default Monoid for bool types


gusty:
The same way it was decided for pragmatic reasons, to use addition as the default monoid for numeric types, now the question is does it makes sense to use a default for monoids?

If yes, should it be the 'or' operation?
This way the `Any` wrapper is no longer needed.

Thoughts ?

gusty:
After exchanging opinions in the FP chat, we came to the conclusion that `XOR` would be a better default.

gusty:
Closed in #238

PR #238, opened 2020-01-15
Use XOR as default monoid for bool

gusty:
This is a proposal of #237


wallymathieu:
Looks simple enough. The one thing is the Any type alias looks a bit weird. Perhaps due to some typescript programming lately.

gusty:
Yes, it's a bit weird.
It provides backwards compatibility of source code.
Probably not at binary level, so maybe this should be rolled out in v2.0

Issue #243, opened 2020-01-17
Module build failed (from ./node_modules/fable-loader/index.js): Error: File src/App.fs cannot be found in source list


entyo:
### Description

Checking Release Notes, I noticed that Fable support of FSharpPlus has been enhanced in recent versions. So I tried to use `1.1.0-RC1` in fable-loader project.

I just wrote some code in `src/App.fs`:

```fsharp
module App

open Fable.Core.JS
open FSharpPlus

let res1 = map ((+) 2) (Some 2)

console.log "hello"
```

But when I try to transpile, I got the error in the title.

```
ERROR in ./src/App.fs
Module build failed (from ./node_modules/fable-loader/index.js):
Error: File /home/e_ntyo/fable-fsharpplus-sample/src/App.fs cannot be found in source list:
/home/e_ntyo/fable-fsharpplus-sample/.fable/FSharpPlus.1.1.0-RC1/Internals.fs
/home/e_ntyo/fable-fsharpplus-sample/.fable/FSharpPlus.1.1.0-RC1/Extensions.fs
/home/e_ntyo/fable-fsharpplus-sample/.fable/FSharpPlus.1.1.0-RC1/Converter.fs
/home/e_ntyo/fable-fsharpplus-sample/.fable/FSharpPlus.1.1.0-RC1/Numeric.fs
/home/e_ntyo/fable-fsharpplus-sample/.fable/FSharpPlus.1.1.0-RC1/Monoid.fs
/home/e_ntyo/fable-fsharpplus-sample/.fable/FSharpPlus.1.1.0-RC1/Functor.fs
/home/e_ntyo/fable-fsharpplus-sample/.fable/FSharpPlus.1.1.0-RC1/Foldable.fs
/home/e_ntyo/fable-fsharpplus-sample/.fable/FSharpPlus.1.1.0-RC1/Bifoldable.fs
/home/e_ntyo/fable-fsharpplus-sample/.fable/FSharpPlus.1.1.0-RC1/Traversable.fs
/home/e_ntyo/fable-fsharpplus-sample/.fable/FSharpPlus.1.1.0-RC1/Indexable.fs
/home/e_ntyo/fable-fsharpplus-sample/.fable/FSharpPlus.1.1.0-RC1/MonadTrans.fs
/home/e_ntyo/fable-fsharpplus-sample/.fable/FSharpPlus.1.1.0-RC1/Collection.fs
[...]

PR #244, opened 2020-01-18
Reproduce issue around operator

wallymathieu:
See https://github.com/fsprojects/FSharpPlus/issues/243
Though it could be an issue with other operators 
Anyone wanting to look further into this is more than welcome.
So far what I've found is that when the operator `map` is used, the GeneralTests file does not get emitted as a javascript file. In order to get full output you can do:
```npm run pretest --verbose ; npm run test --verbose```
in the FSharpPlusFable.Tests folder


gusty:
Closed by #283

PR #247, opened 2020-01-22
Minor tweaks

adz:
Some doc tweaks...

Also has dotnet tool manifest file, and a script I can run until I can get build.fsx to work.
Used very latest paket, to get full dotnetcore support, but can downgrade if problematic.

Will be used only if you use `dotnet tool restore` and `dotnet paket`


manually_build_docs.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+# Restore fake and paket
+dotnet tool restore
+
+# Restore paket deps
+dotnet paket restore
```
wallymathieu: Nice! Perhaps something to have instead of the existing bootstrapped?

manually_build_docs.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
```
wallymathieu: You could put this in the docsrc folder in order for it not to be in the root.

.paket/Paket.Restore.targets
```diff
@@ -5,6 +5,11 @@
     <!-- make MSBuild track this file for incremental builds. -->
```
wallymathieu: Perhaps we could remove this file altogether since the projects are not dependant on paket to load

manually_build_docs.sh
adz: Good idea. I didn't really want to commit it as fake build should be enough, but doesn't work for me right now. It still has value to show simpler steps, bit should eventually be removed.

manually_build_docs.sh
adz: That'd be better... want to focus on bootstrapping separately though...

manually_build_docs.sh
adz: Done, thanks!

.paket/Paket.Restore.targets
adz: Deleted it, and added to gitignore

PR #248, opened 2020-01-23
Deprecate some functions in Validation

wallymathieu:


src/FSharpPlus/Validations.fs
```diff
@@ -113,6 +113,7 @@ module Validation=
     ///
     /// validate : 'e -> ('a -> bool) -> 'a -> Validation<'e, 'a>
     ///
+    [<System.Obsolete("The validate method is less generic than intended.")>]
```
gusty: I don't think we should give that reason for the deprecation. Let's put something like "This function will not be supported in future versions."

gusty:
We can also take the opportunity to rename (and deprecate) both `orElse`  and `valueOr` as they should be aligned with the recent naming (which at the same time is aligned with the existing F# core functions) `defaultValue` and `defaultWith`.

src/FSharpPlus/Validations.fs
wallymathieu: Let's go with that then

src/FSharpPlus/Validations.fs
```diff
@@ -77,12 +77,14 @@ module Validation=
...
 
     /// orElse v a returns 'a when v is Failure, and the a in Success a.
+    [<System.Obsolete("This function will not be supported in future versions.")>]
     let orElse v (a: 'a) = 
         match v with
         | Failure _ -> a
         | Success x -> x
 
     /// Return the 'a or run the given function over the 'e.
+    [<System.Obsolete("This function will not be supported in future versions.")>]
```
gusty: Great, but for these functions the reason is not the same. We can put here "Use Validation.defaultWith instead."

wallymathieu: We should add the `defaultWith` and `defaultValue` functions as well

gusty:
All green. Please Squash and Merge at your convenience.

wallymathieu:
Thanks @gusty for the help!

PR #252, opened 2020-01-24
Deprecate non-generic catch

gusty:
Non generic `catch` functions have the arguments flipped with respect to the generic version.

This could be useful for using forward pipe expressions, but now we have `defaultWith` for that in Result, so this function is redundant. It would be better to have it with the arguments the same way as the generic.

For `Choice` we can add a function, maybe something like `bind2Of2`.

### DIFF
--- a/src/FSharpPlus/Extensions.fs
+++ b/src/FSharpPlus/Extensions.fs
@@ -62,7 +62,10 @@ module Result =
     let throw  x = Error x
     let apply f x = match f, x with   Ok a, Ok b      -> Ok (a b) | Error e, _ | _, Error e -> Error e      : Result<'b,'e>
     let flatten                  = function Ok (Ok v) -> Ok v     | Ok (Error e)  | Error e -> Error e
+    
+    [<System.Obsolete("Use Result.defaultWith instead.")>]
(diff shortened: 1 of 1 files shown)


gusty:
> Does obsolete show up in docs?

Good point, let's generate them and see.

gusty:
I tried look at the obsolete methods and found out the `Validation` module and `Result` for some dark reason are not in the reference.

Choice is, and there I still see `catch`, it looks like it doesn't care about the obsolete attribute. I think we should exclude them from the reference.

gusty:
Well, it's not that it doesn't care. Is rather than API reference stopped working long time ago !

I think I managed to get it back with something I have on the drafts.

From what I see now it updated the docs, and YES, it shows the deprecated function with a message.

adz:
Oh I hadn't been able to gen API docs ref yet. If you can relate what you did I'd appreciate it!

That looks good.

gusty:
I'll create a PR shortly, so you can see it.

I removed almost all build stuff, and used standard dotnet for build and a small script to create the docs.

No more fake and paket needed !

wallymathieu:
So much simpler then 😄

PR #253, opened 2020-01-24
Reword 1st part of the dev guide

gusty:
An attempt to clarify the ideas expressed there.

### DIFF
--- a/DEVELOPER_GUIDE.md
+++ b/DEVELOPER_GUIDE.md
@@ -1,6 +1,13 @@
 # DEVELOPER GUIDE
 
+### Extensions
 
+This library defines many extensions for different types.
+Some of these functions are conceptually connected, in that they represent an abstraction and sometimes they have indeed a generic version of the function which doesn't require typing the type as module prefix.
+Note that in these cases, normally the names are the same but this is not a rule, there are cases where the name differs to avoid collisions or confusion with other functions, for example:
+
+ - `map` is a generic function, and we have many types implementing it, but for dictionaries this correspond to `mapValues`
+ - `min` is a non-generic function operating in collections, but its generic counterpart is `minimum` to avoid collision with the built-in `min` function (minimum between two values).
 
 ### Definition of abstraction
 
@@ -22,7 +29,7 @@ Those Invokers contains static methods of several kinds:
 * an _invokable_ method, that should always be defined as `inline`, which embeds the SRTP dispatcher machinery that is used to pick the right overload (whether defined in F#+ or outside of it)
 * a default or set of default implementations, those may need to be defined in an instrinsic type extension due to F# compiler implementation details to enable being picked up correctly for types that end up supporting the abstraction when such abstraction comes with a default implementation expressed in terms of composing several functions, those are always defined as `inline`.
 
-### Implementation idioms
+
 
 #### Concrete implementations
 
--- a/DEVELOPER_GUIDE.md
+++ b/DEVELOPER_GUIDE.md
@@ -8,6 +8,22 @@ Note that in these cases, normally the names are the same but this is not a rule
 
  - `map` is a generic function, and we have many types implementing it, but for dictionaries this correspond to `mapValues`
  - `min` is a non-generic function operating in collections, but its generic counterpart is `minimum` to avoid collision with the built-in `min` function (minimum between two values).
+ 
+So, what this means is, we're not providing generic functions based on names, although in many cases names are the same, we need to take into account:
+ 
+ - F# core has some inconsistencies as it's not a type classes based library. A typeclasses based library is not necessarily something that implments a trick for typeclasses, but something that is designed as if we had support for them, I mean capturing some generic concepts and making them clear in the names choosen for every functions.
+
+ - But here we're not attempting to fix F# core, we want to build on top in a non intrusive way, and try to re-use all the consistent concepts, idioms and de-facto naming conventions from F# core as much as possible without increasing the inconsistency level already there.
+
+So the solution sometimes require some creativity like thinking new names that make it clear what the function does without departing too far from existing naming conventions. 
+ 
+Another interesting case is the `zip` related functions:
+
+ - For collections like types, in F# core it's normal to find `zip` / `map2` functions, which acts pairwise. But another possible implementation is the applicative zip which works cross product.
+ - Here we define 2 generic functions, `lift2` which corresponds always to the applicative instance, so normally in non-collections it will correspond to the non-generic `map2` but since collections already have (or at least it's expetcted to) a `map2` acting pairwise, in those cases only we provide a `lift2` extension and that's what's used for its generic counterpart.
+ 
+ - The other generic function we provide is `zip` which is available mainly for collection like types and although they match the non-generic name, note that the behavior is not exactly the same, because F# core throws errors for list and arrays when the number of elements are different. So we could say that in fact it matches `.zipShortest` when defined, otherwise it matchs `.zip`.
+
 
 ### Definition of abstraction
 
--- a/DEVELOPER_GUIDE.md
+++ b/DEVELOPER_GUIDE.md
@@ -2,23 +2,25 @@
 
 
 
-### Definition of a type class
+### Definition of abstraction
 
-The concept of type class is a type system construct enabling ad-hoc polymorphism.
+The abstractions represented in this library are a set of types that in fact represents generic functions, we call these types and we'll refer to as "Invokables".
 
-in F#+, due to the constraint of targetting CIL, which doesn't have first class support for type class definition and implementation, those are not defined as F# types; instead, they are referred to in generic type parameters in the various type signatures that forms the idioms of this library implementation.
+This Invokables are organized in such a way that resemble typeclasses, concepts or traits in other languages where they have a type system construct enabling higher kinds and ad-hoc polymorphism.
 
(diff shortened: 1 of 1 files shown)


adz:
Could you add a paragraph explaining SRTP? I had no idea it existed when first finding this library, and am still a little unclear about which overload gets chosen with respect to the default1, 2, etc... last arg.

DEVELOPER_GUIDE.md
```diff
@@ -2,23 +2,25 @@
...
 
-Due to that, they don't carry any effect on implementation as they are just arbitrary labels for the generic type parameters, useful to the developers and users of the library.
+In F#+, due to the constraint of targetting CIL, which doesn't have first class support for Higher Kinds, we can't enforce that a specific type parameter belongs to a specific abstraction. So generic type parameters are normally given the name of the abstraction, in the various type signatures that forms the idioms of this library implementation.
 
-The concrete implementation of them on the other hand is done, minimally, by defining a concrete type per function.
+Due to that, they don't carry any effect on implementation as they are just arbitrary labels for the generic type parameters, useful to communicate the intention to the developers and users of the library. Still F# type system will normall show some constraints at the method level, but these constraints are sometimes not very helpful in communicating the related abstraction.
```
wallymathieu: *normall -> normally

DEVELOPER_GUIDE.md
```diff
@@ -2,23 +2,25 @@
...
-* an _invoker_ method, that should always be defined as `inline`, which embeds the SRTP machinery that is used to pick the right overload (whether defined in F#+ or outside of it for implementing type class)
-* a default or set of default implementations, those may need to be defined in a type extension due to F# compiler implementation details to enable being picked up correctly for types that end up supporting the type class when such type class come with default implementation expressed in terms of composing several functions, those are also always defined as `inline`.
+* an _invokable_ method, that should always be defined as `inline`, which embeds the SRTP machinery that is used to pick the right overload (whether defined in F#+ or outside of it for implementing type class)
+* a default or set of default implementations, those may need to be defined in an instrinsec type extension due to F# compiler implementation details to enable being picked up correctly for types that end up supporting the type class when such type class come with default implementation expressed in terms of composing several functions, those are also always defined as `inline`.
```
wallymathieu: *instrinsec -> intrinsic ?

DEVELOPER_GUIDE.md
```diff
@@ -1,29 +1,44 @@
...
 
-* concrete overloads, for exemple Monoid's `Plus` will have overloads of a `Plus` method (named the same as the type defining those) for each of the BCL and FSharp.Core types that ought to support the `Monoid` type class, e.g. one to concatenate two strings, one to concatenate two lists, one to concatenate two arrays, and so on and so forth.
-* an _invoker_ method, that should always be defined as `inline`, which embeds the SRTP machinery that is used to pick the right overload (whether defined in F#+ or outside of it for implementing type class)
-* a default or set of default implementations, those may need to be defined in a type extension due to F# compiler implementation details to enable being picked up correctly for types that end up supporting the type class when such type class come with default implementation expressed in terms of composing several functions, those are also always defined as `inline`.
 
-### Implementation idioms
 
 #### Concrete implementations
 
+We should add as much concrete implementations for primitive types as possible, since the end user won't be able to add them later.
```
3Rafal, line 36: I wonder what `primitive types` means in this context. 
Googling "F# primitive types" returns docs: https://docs.microsoft.com/pl-pl/dotnet/fsharp/language-reference/basic-types
I guess that what you've meant here was FSharp.Core types, but I may be wrong.

DEVELOPER_GUIDE.md
```diff
@@ -1,29 +1,44 @@
 # DEVELOPER GUIDE
 
+### Extensions
 
+This library defines many extensions for different types.
+Some of these functions are conceptually connected, in that they represent an abstraction and sometimes they have indeed a generic version of the function which doesn't require typing the type as module prefix.
+Note that in these cases, normally the names are the same but this is not a rule, there are cases where the name differs to avoid collisions or confusion with other functions, for example:
```
3Rafal, line 7: You gave a very interesting explanation of relation of F#+ and F# core library on slack. I can't search it, because I don't have premium Slack version, but it went something along the lines of:
"In F#+ we try to expand FSharp.Core functionalities, but we are trying not to be intrusive"
There is a remark on this in general readme
>F#+ builds upon FSharp, using generic programming techniques to help you avoid boiler plate code. However, by naming conventions and signatures it can be seen to 'enhance' rather than 'replace' existing patterns as much as possible.

But I wonder if we could refer to this concept in a more concrete and contributor-oriented way here.

DEVELOPER_GUIDE.md
```diff
@@ -1,29 +1,44 @@
...
 
-### Definition of a type class
+ - `map` is a generic function, and we have many types implementing it, but for dictionaries this correspond to `mapValues`
```
3Rafal, line 9: I wonder if we could add a zip case, as it is very interesting one. We have safe behavior in generic function, that "fixes" inconsistencies between Seq.zip and List.zip, Array.zip in FSharp.Core. I think this is a good example of F#+ approach of generalizing functions with consistent and predictable behavior.

DEVELOPER_GUIDE.md
```diff
@@ -1,29 +1,44 @@
...
+The abstractions represented in the Control namespace of this library are a set of types that in fact represents generic functions, we call these types and we'll refer to as "Invokables".
 
-Due to that, they don't carry any effect on implementation as they are just arbitrary labels for the generic type parameters, useful to the developers and users of the library.
+These Invokables are organized in such a way that resemble typeclasses, concepts or traits in other languages where they have a type system construct enabling higher kinds and ad-hoc polymorphism.
 
-The concrete implementation of them on the other hand is done, minimally, by defining a concrete type per function.
+In F#+, due to the constraint of targeting CIL which doesn't have first class support for Higher Kinds, we can't enforce that a specific type parameter belongs to a specific abstraction. So generic type parameters are normally given the name of the abstraction, in the various type signatures that forms the idioms of this library implementation.
```
3Rafal, line 18: I wonder if we could add code example for named generic parameter. Discussed concept are highly abstract, and it could help make a connection between high-level concepts and their implementation in F#+ codebase

DEVELOPER_GUIDE.md
```diff
@@ -1,29 +1,44 @@
...
 
-### Implementation idioms
 
 #### Concrete implementations
 
+We should add as much concrete implementations for primitive types as possible, since the end user won't be able to add them later.
+
 #### Defaults
 
+The goal of defaults is to allow users of the library to write less code, but their are not intended to be used by developers of F#+ as we should afford writing more code in order to maximize usability of the library.
```
3Rafal, line 40: We describe a goal of defaults, but it's not clear what are they. Do you mean "a default or set of default implementations" which were described above?
I'm also confused about how they are not intended to be used by developers?
Or do we want to say that we try to make a library which principle is to allow user to write less boilerplate, and because of that we have to put some boilerplate inside F#+?

3Rafal:
@gusty , it's great that you're enhancing Developer's Guide. I think it will ease the learning curve for new contributors.
I have a general sense, that ideas that you're describing are quite abstract, especially for a newcomer. I wonder if we could make it easier by using code examples/snippets. They would obviously have to be quite generic and high-level. But I guess it could be worth it. 
I think it's a similar case to your blog posts. Topics you've described were highly technical, but you were using code examples and it was easy to follow.

DEVELOPER_GUIDE.md
gusty: In fact I mean both .NET framework types and F# core types, which are the sole dependencies of this library. Not sure what's the best word for describing them.

DEVELOPER_GUIDE.md
gusty: Yes, it's no longer available. But I think what I said was basically:

DEVELOPER_GUIDE.md
gusty: Good point

DEVELOPER_GUIDE.md
gusty: Definitely, we should add snippets otherwise it's too blah-blah guide, feel free to suggest concrete code fragments to include here.

DEVELOPER_GUIDE.md
gusty: Yes, it's rather  the latter and it's not clear at all, I wrote it this way mainly for me to not forget to include this idea in the guideline, but need to come back later and do the effort of explaining it in a better way, ideally with some concrete examples.

DEVELOPER_GUIDE.md
3Rafal: I think you've already referenced them in this guide as `BCL and FSharp.Core types`, which I find accurate.

DEVELOPER_GUIDE.md
3Rafal: I think, that information from above post should be put inside Developers Guide. 
Currently, dev guide consist mostly of implementation explanation. I think it is a good idea to show high-level "philosophy of F#+" and it's relation to core F# library.  It would make it easier for a newcomer to be aligned with the purpose of this project. It would also give a direction on how we should expand it.

gusty:
I added more content and tried to address all valuable feedback (thanks @3Rafal , we miss you).
If there are no objections I would like to merge it.

PR #257, opened 2020-01-28
Split up Extensions.fs into file per module

adz:
Note there is Extensions/Extensions.fs which holds the actual module called "Extensions" which is opened to get extensions on types, rather than 'additional operations' that come in via modules.

I suppose technically, these modules are not really extensions?
They are just modules with functions?


gusty:
> I suppose technically, these modules are not really extensions?
> They are just modules with functions?

Yes they are.
But now I think we can group them in the same file.
I mean for instance in Task.fs the module with all the functions but also the Extension members.
What do you think?

adz:
Right now, you have to "open FSharpData.Extensions" to get what is in Extensions.fs because it's under `module Extensions`.

We can't add the same module under each of List.fs and Task.fs since they share the same namespace `FSharpData` (i.e. no sub-namespace).

So, we could change behaviour such that you get these type extensions (currently in Extensions.fs) without having to open?

Not sure if there's another idea

adz:
Oops, I means "FSharpPlus" not data!

I like the idea that you open at a 'deeper' level if you don't want it all...
So if you open "FSharpPlus" it auto-opens "FSharpPlus.Extensions" and all these move under that namespace which is AutoOpen-ed.

I'd do the same with Operators, but with another level "FSharpPlus.Operators.CommonOperators" where if you want them all, you open FShapPlus or FSharpPlus.Operators, or you can open smaller context...

I don't know that it's so important, but it helps the user understand what they are getting..
WDYT?

I know this is 2.0 stuff.... so could leave as is, and for 2.0 introduce additional namespacing...

gusty:
I see. Well then we can leave it as you did so far and for v2 we can do another PR which spread extensions into corresponding files and does the autoopen thing.

PR #262, opened 2020-01-30
adjust other flags

dsyme:
This makes it possible to set additional OtherFlags in Directory.Build.props


wallymathieu:
Thanks for the contribution @dsyme !

PR #263, opened 2020-02-02
Lens: add _at and non for working with Maps

wuzzeb:
The existing Map._item isn't a nice lens to work with since the getter/view returns type Option<'a> but the setter receives 'a.  That is, it violates the lens law of setting what you get doesn't change anything `set l (view l s) s ~ s` (It doesn't even compile.)

I added the _at lens https://hackage.haskell.org/package/lens-4.18.1/docs/Control-Lens-At.html#t:At which both the getter and setter use Option<'a>.

I also added non (https://hackage.haskell.org/package/lens-4.18.1/docs/Control-Lens-Iso.html#v:non) which works well together with _at.

### DIFF
--- a/src/FSharpPlus/Lens.fs
+++ b/src/FSharpPlus/Lens.fs
@@ -116,9 +116,15 @@ module Lens =
     [<RequireQualifiedAccess>]
     module Map=
         let inline _item i f t = Map.InvokeOnInstance (fun x -> Map.add i x t) (f (Map.tryFind i t))
+        let inline _at i f t = Map.InvokeOnInstance
+                                  (function | None -> Map.remove i t | Some(x) -> Map.add i x t)
+                                  (f (Map.tryFind i t))
     [<RequireQualifiedAccess>]
     module IReadOnlyDictionary=
         let inline _item i f t = Map.InvokeOnInstance (fun x -> IReadOnlyDictionary.add i x t) (f (IReadOnlyDictionary.tryGetValue i t))
+
+    /// Lens for the value inside an Option or the given default value if the Option is None.  Works well when combined with Map._at
+    let inline non def f ma = Map.InvokeOnInstance (fun a' -> if a' = def then None else Some(a')) (f (Option.defaultValue def ma))
 
     // Prism
 
--- a/src/FSharpPlus/Lens.fs
+++ b/src/FSharpPlus/Lens.fs
@@ -115,15 +115,22 @@ module Lens =
 
     [<RequireQualifiedAccess>]
     module Map=
-        let inline _item i f t = Map.InvokeOnInstance (fun x -> Map.add i x t) (f (Map.tryFind i t))
-        let inline _at i f t = Map.InvokeOnInstance
+        /// Given a specific key, produces a Lens from a Map<key, value> to an Option<value>.  When setting,
+        /// a Some(value) will insert or replace the value into the map at the given key.  Setting a value of
+        /// None will delete the value at the specified key.  Works well together with non.
+        let inline _item i f t = Map.InvokeOnInstance
                                   (function | None -> Map.remove i t | Some(x) -> Map.add i x t)
                                   (f (Map.tryFind i t))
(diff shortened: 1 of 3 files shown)


src/FSharpPlus/Lens.fs
```diff
@@ -116,10 +116,16 @@ module Lens =
     [<RequireQualifiedAccess>]
     module Map=
         let inline _item i f t = Map.InvokeOnInstance (fun x -> Map.add i x t) (f (Map.tryFind i t))
+        let inline _at i f t = Map.InvokeOnInstance
```
wallymathieu: I'm unsure. Perhaps it would be better to redefine `_item` instead of having two slightly similar functions?

wuzzeb: I agree, _at can do everything _item does and more when combined with non. But don't know how much you care about backwards compatibility... changing _item would produce a compile error everywhere it is used.  Also, the compile error is pretty non-helpful since it says something about obj can't be inferred.

wallymathieu: `_item` is not in the 1.0.0 release so it would probably be better to fix the signature of that function for `Map` and `ReadOnlyDictionary`

wallymathieu: Since otherwise there would be a need to fix the issue in a 2.0.0 release.

gusty:
I agree with @wallymathieu  in that if something is fundamentally wrong and it's only in the CIs we can still fix it, noting it in the change log.

But I also wonder, if this function is called `_at` in Haskell, shouldn't we deprecate `_item` and use this name instead?

Otherwise it could be confusing for people familiar with the Haskell implementation if we switch names.

@wuzzeb  please let me know if this is the case, as I'm a bit lost with `_item` and `_at` in the original implementation.

wuzzeb:
In Haskell, there are two basic lenses/Transversals for maps: at and ix

* at, given a key, produces a lens from Map<key, value> to Option<value> which is what I implemented above.  Setting None causes the value to be deleted, setting Some(value) adds or replaces the item.

https://hackage.haskell.org/package/microlens-0.4.11.2/docs/Lens-Micro.html#v:at
https://hackage.haskell.org/package/lens-4.18.1/docs/Control-Lens-At.html#t:At

(Side note, I like the microlens documentation much better for learning, the microlens at shows the use of at in a simple manor)

* ix is a Transversal (lens where the functor is an applicative functor).  That is, given a key it produces a (Applicative f) => Map<key, value> -> f (Map <key, value>)

https://hackage.haskell.org/package/microlens-0.4.11.2/docs/Lens-Micro.html#v:ix

* The existing FSharpPlus _item doesn't have an equivalent in Haskell and I don't even think it is a lens.  The getter gets an Option<value> but the setter only takes a value.

gusty:
@wuzzeb thanks for the detailed explanation.

If that's the case I would say, we can safely deprecate `_item` and add the new function as `_at`.
This way:

 - We'll be fixing the issue
 - We'll be adding a function with right name
 - We won't break anyone, even from the CIs

The error messages are bad, there's no much we can do about it, let's hope new improvements in the compiler make things better.

gusty:
The only thing could make us change the name is when there is already a different naming convention in F#.

For instance, in this lib, for standard functions we use `nth` to refer to number based indices and `item` to generic ones, which could be keys to dictionaries. Not sure if `ix` and `at` comes from normal Haskell functions.

wallymathieu:
I think I added `_item` to match the naming of `item` in f#+ , you can find the PR here: https://github.com/fsprojects/FSharpPlus/pull/143
The functions that influenced the naming can be found here:
https://github.com/fsprojects/FSharpPlus/blob/8e48fbfd6c834e3f9ef451f9b16ab8bcfcc80c50/src/FSharpPlus/Operators.fs#L404-L408
So the question is in what direction we should go?

wallymathieu:
Thanks for the contribution @wuzzeb !

Issue #234, opened 2020-01-14
Rename liftA2 to lift2


gusty:
I think `liftA2` is here for historical reasons: we used to have `lift2M` as well.

I don't think we need to keep calling it `liftA2`, we can rename it to `lift2` and deprecate the old name.

MNie:
@gusty I think we could close this one since pr was merged ;)

gusty:
Sure, thanks for the reminder.

Closed as per #234

PR #264, opened 2020-02-04
#234 rename lifta2 to lift2

MNie:
Add liftA function and replace all of the usages of liftA2.

Related: #234

### DIFF
--- a/src/FSharpPlus/Operators.fs
+++ b/src/FSharpPlus/Operators.fs
@@ -97,7 +97,11 @@ module Operators =
     let inline (<*>) (f: '``Applicative<'T -> 'U>``) (x: '``Applicative<'T>``) : '``Applicative<'U>`` = Apply.Invoke f x : '``Applicative<'U>``
 
     /// Apply 2 lifted arguments to a non-lifted function.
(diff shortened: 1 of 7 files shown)


src/FSharpPlus/Operators.fs
```diff
@@ -97,7 +97,11 @@ module Operators =
     let inline (<*>) (f: '``Applicative<'T -> 'U>``) (x: '``Applicative<'T>``) : '``Applicative<'U>`` = Apply.Invoke f x : '``Applicative<'U>``
 
     /// Apply 2 lifted arguments to a non-lifted function.
-    let inline liftA2 (f: 'T->'U->'V) (x: '``Applicative<'T>``) (y: '``Applicative<'U>``) : '``Applicative<'V>`` = (f <!> x : '``Applicative<'U->'V>``) <*> y
```
gusty: If you simply remove `liftA2` it would be a breaking change.

Could you deprecate it instead? As we did with other functions, like `valueOr` in Validation.fs

MNie: It was moved to line 104 so we could use inside a `lift2` function without duplicating the logic inside.

gusty: Indeed, sorry, my bad.

PR #265, opened 2020-02-05
Document extensions

adz:
Still WIP, but want to put this up for discussion.

The idea was to more conversationally introduce extensions from perspective of utility to user, rather than list off like an API doc.

I dumped all the functions into the doc, which was useful to 'see' them all, but not sure they'll stay -- can just link to the appropriate API doc. However, might still be useful to group within a module (eg map, apply, bind together) instead of alphabetical ordering default of API.

### DIFF
--- a/docsrc/content/extensions.fsx
+++ b/docsrc/content/extensions.fsx
@@ -21,18 +21,20 @@ such as `map`, `bind` and `apply` on List, Array, and Seq, but also Option and R
 
 Construction:
 =============
-This is added and is already defined for Seq, Array and List:
+The `singleton` function is already defined for Seq, Array and List, but F#+ adds it for Enumerator:
 
  * Enumerator.singleton - construct a container with the given value inside it
 
 To construct MonadError instances (Result or Choice) you can use result/throw:
 
- * result - construct with the given value (as Ok or Choice1Of2)
- * throw - construct an error from the given value (as Error or Choice2of2)
+ * Result.result / Choice.result - construct with the given value (as Ok or Choice1Of2)
+ * Result.throw / Choice.throw - construct an error from the given value (as Error or Choice2of2)
 
 It's also possible to construct by wrapping exception producing functions:
 
- * protect - protect a function in try/catch, returning a wrapped value
+ * Option.protect - returns None on exception
+ * Result.protect - returns Error with exception value on exception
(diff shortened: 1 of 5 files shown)


docsrc/content/extensions.fsx
```diff
@@ -0,0 +1,295 @@
...
+=======================
+Extensions are what you probably expect: helper functions for existing types.
+ 
+They are defined as modules with the same name as the types they operate on
+under the FSharpPlus namespace, so can be accessed via:
+*)
+open FSharpPlus
+
+(**
+Some functions are common across traversable types such as `intercalate` on
```
adz: foldable probably here

docsrc/content/extensions.fsx
```diff
@@ -0,0 +1,295 @@
...
+
+ * either - unwraps the result/choice by applying the given `ok` or `err` function as appropriate
+
+Note that there is also the generic `either` operator function that works
+exactly the same as `Result.either`.
+
+Also, see the generic function [`option`](reference/fsharpplus-operators.html) that
+unwraps an Option in a similar way to `either`.
+
+On Foldables (collections)
```
gusty: Note that foldable and collections are actually two different abstractions although they are related.
A collection in some sense is like the opposite to a foldable.

gusty:
Looks good in general.

Can you deploy it so we review it online. It will never be worst than nothing which what we have currently ;)

docsrc/content/extensions.fsx
adz: Hmmm, in what sense do you mean a collection is opposite?

gusty: Well, a foldable is something that can reduce to a single value. And a Collection is something that can unfold to many values.

That's why we use toSeq for Foldable and OfSeq for collections.

It's not very clear this explanation as it's not really clear the distinction, specially of the latter, so just take as a general guide.

adz: I have never thought of a sequence like an unfoldable, but of course it is :)

Thanks.

adz:
Not sure why the menu to API was commented out https://github.com/fsprojects/FSharpPlus/blob/master/docsrc/tools/templates/template.fsx#L90

I re-enabled it and it seems to work fine.

wallymathieu:
The API generation didn't work. We have forgotten to add the link again after fixing it.

Issue #41, opened 2017-03-05
Traversing infinite sequences


gusty:
_From @mausch on April 16, 2014 21:23_

Traversing an infinite list works fine in Haskell: http://ideone.com/GuytMd
But the equivalent F# + FSharpPlus code doesn't terminate and eats infinite memory:

``` fsharp
let x = traverse (fun x -> if x > 4 then Some x else None) (Seq.initInfinite id)
```

This is probably because [foldr is currently strict for seq](https://github.com/gmpl/FsControl/blob/master/FsControl.Core/Foldable.fs#L21)

Maybe a specialized foldr would work here, e.g. http://fpish.net/blog/anton.tayanovskyy/id/1453/http~3a~2f~2fwww.intellifactory.com~2fblogs~2fanton.tayanovskyy~2f2009~2f12~2f11~2fFoldr-or-FoldBack-on-Infinite-F~21sharp~21-Sequences.article

(BTW [this stackoverflow question](http://stackoverflow.com/q/23098484/21239) made me realize this).

_Copied from original issue: gmpl/FsControl#26_

gmpl:
I'm not sure if is possible to define a <code>foldr</code> working over infinite sequences. I would rather try to write a more efficient <code>traverse</code> for seq using low level enumerators instead of <code>foldr</code>. Still I'm not sure if it would be possible to make it work with infinite sequences, but I would give it a try.

gmpl:
_From @mausch on April 17, 2014 22:49_

Yes, such a foldr is not possible with the usual signature, so yes, the 'optimization' needs to happen in the implementation of traverse. I'm currently trying to implement such a traverse for a LazyList, then the implementation for seq would just internally use the one for LazyList.
Doing this using low-level enumerators as you say should also be possible, it just seems more complicated.

gmpl:
Hopefully I'm wrong but I'm afraid that even implementing a traverse for <code>LazyList</code> will not be lazy enough. A poor man solution is to add specific instances, I did one that is able to run your sample code. https://github.com/gmpl/FsControl/commit/0496a739c20b8fdf57cbb9a91e382b13dac07459

gmpl:
The poor man solution seems to be the way to approach this problem.
Each time an optimized or a lazy version is needed a specific instance should be added.
I would say there will be more cases like this, in which the lazyness is not the default since it will not be automatically derived, see for instance the new ``Delay`` method for workflows.

gmpl:
[...]

PR #268, opened 2020-02-07
Generalize short-circuit for Traverse

gusty:
This PR will offer a general solution for #41 

The solution presented here adds a (redundant) method to `Apply` intended for internal use, called `IsLeftZeroForApply`.

The theory behind it is that a binary operation can have one (or more) left-zero elements, a left zero element is a element `z` such that:

    z * e = z

So, we add overloads of this method for some types that we can tell about their left-zero elements. But for types that are Alternatives we don't need an overload because:

    empty <*> f = empty

Which means we can use that knowledge automatically.

For types we can't tell about their left-zero elements we don't add an overload.

### DIFF
--- a/src/FSharpPlus/Control/Traversable.fs
+++ b/src/FSharpPlus/Control/Traversable.fs
@@ -16,7 +16,7 @@ type Sequence =
 
     #if !FABLE_COMPILER
     [<EditorBrowsable(EditorBrowsableState.Never)>]
-    static member inline ForInfiniteSequences (t: seq<_>, isFailure) =
+    static member inline ForInfiniteSequences (t: seq<_>, isFailure, conversion) =
         let add x y = Array.append x [|y|]
         let mutable go = true
         let mutable r = result [||]
@@ -24,7 +24,7 @@ type Sequence =
         while e.MoveNext () && go do
             if isFailure e.Current then go <- false
             r <- Map.Invoke add r <*> e.Current
-        Map.Invoke Array.toSeq r
+        Map.Invoke conversion r
     #endif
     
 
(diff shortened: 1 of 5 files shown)


src/FSharpPlus/Control/Traversable.fs
```diff
@@ -49,15 +47,9 @@ type Traverse =
...
-       let res = Seq.toArray (seq {
-           use e = t.GetEnumerator ()
-           while e.MoveNext() && ok.Value do
-               match f e.Current with
-               | Some v -> yield v
-               | None   -> ok.Value <- false})
-       if ok.Value then Some (Array.toSeq res) else None
+    static member inline Traverse (t: seq<'T>, f: 'T->'``Functor<'U>``, [<Optional>]_output: '``Functor<seq<'U>>``, [<Optional>]_impl: Default2) =
+        let mapped = Seq.map f t
+        Sequence.ForInfiniteSequences (mapped, IsLeftZeroForApply.Invoke) : '``Functor<seq<'U>>``
```
wallymathieu: The `Sequence.ForInfiniteSequences` looks a bit odd now that I'm looking at it. Though it seems to do the same thing as the left side. Why `ForInfiniteSequences` when it does to array?

tests/FSharpPlus.Tests/General.fs
```diff
@@ -907,7 +907,14 @@ module Monad =
 #endif
 
 
-module Traversable = 
+module Traversable =
+
+    type Either<'l,'r> = Left of 'l | Right of 'r with
```
wallymathieu: 👍

src/FSharpPlus/Control/Traversable.fs
```diff
@@ -17,16 +17,13 @@ type Sequence =
...
-        let cons x y = Array.append [|x|] y
-        let cons_f x ys = Map.Invoke cons x <*> ys
-        let r = Array.foldBack cons_f buf (result [||])
+        let add x y = Array.append x [|y|]
+        let mutable go = true
+        let mutable r = result [||]
+        use e = t.GetEnumerator ()
+        while e.MoveNext () && go do
+            if isFailure e.Current then go <- false
+            r <- Map.Invoke add r <*> e.Current
```
wallymathieu: Nice!

PR #271, opened 2020-02-14
Fixes un/curry to allow arguments of different types

NickDarvey:
Regression in https://github.com/fsprojects/FSharpPlus/commit/804cf417a34328e7dc7fb2b37bf00479d54353ab


src/FSharpPlus/Operators.fs
```diff
@@ -18,13 +18,13 @@ module Operators =
     let inline konst (k: 'T) = fun (_: 'Ignored) -> k
 
     /// Takes a function expecting a tuple of two elements and returns a function expecting two curried arguments.
-    let inline curry f (x: 'T1) (y: 'T1) : 'Result = f (x, y)
+    let inline curry f (x: 'T1) (y: 'T2) : 'Result = f (x, y)
```
wallymathieu: looks like a correct change

PR #274, opened 2020-02-18
Add Windows build with github actions

gusty:
This will add another build.
This one will be with Github's actions.

### DIFF
--- a/.github/workflows/dotnetcore.yml
+++ b/.github/workflows/dotnetcore.yml
@@ -14,4 +14,4 @@ jobs:
       with:
         dotnet-version: 2.2.108
     - name: Build with dotnet
-      run: dotnet build --configuration Release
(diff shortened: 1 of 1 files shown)


.github/workflows/dotnetcore.yml
```diff
@@ -0,0 +1,19 @@
...
+  build:
+
+    runs-on: windows-latest
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Setup .NET Core
+      uses: actions/setup-dotnet@v1
+      with:
+        dotnet-version: 2.2.108
```
gusty: Not sure about the dotnet version.
Maybe we should use the one that it's pinned in the global file.

wallymathieu: There is one version used in .travis.yml as well

PR #275, opened 2020-02-18
exe instead of fsi

wallymathieu:
- [x] make sure that the docs are rendered


wallymathieu:
@adz perhaps you could see if this works for doc gen ?

adz:
@wallymathieu Thanks for your work here! It worked for me on linux, including API docs :raised_hands: 

I did notice there's still a bit of "mono" referenced in `./docsrc/tools/download_nugets.sh` to use nuget.exe. It also worked fine on my mono (6.4).

Is there reason for the mono bit, or is it still in process of changing to dotnet core cli?

wallymathieu:
We can change it to dotnet if that works. So no particular reason other than that it's a common pattern.

PR #276, opened 2020-02-19
Better default for Delay

gusty:
This PR will switch the default for `delay` to the one that it's technically correct for delayed builders.

Auto-sense had to be adjusted because the bind operation inside this delay implementation would fail with a null value.

Also, type inference is a bit more complicate which forced me to add a specific delay for Free<'T> with a trivial implementation since it's not a "delayed monad".

It would be interesting to add more testing and see if there is a change in functionality, good or bad (breaking).

### DIFF
--- a/src/FSharpPlus/Builders.fs
+++ b/src/FSharpPlus/Builders.fs
@@ -119,7 +119,7 @@ module Builders =
             fix ()
         member inline this.For (p: #seq<'T>, rest: 'T->'``MonadPlus<'U>``) : '``MonadPlus<'U>`` =
             let mutable isReallyDelayed = true
-            Delay.Invoke (fun () -> isReallyDelayed <- false; Unchecked.defaultof<'``MonadPlus<'U>``>) |> ignore
+            Delay.Invoke (fun () -> isReallyDelayed <- false; Empty.Invoke () : '``MonadPlus<'U>``) |> ignore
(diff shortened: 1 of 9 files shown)


gusty:
Note that the code broken by this PR is a code that sooner or later would have break as it uses the same type parameter in the declaration of the type and the declaration of the static members, which is a practice we know it doesn't take you too far.

src/FSharpPlus/Builders.fs
```diff
@@ -119,7 +119,7 @@ module Builders =
             fix ()
         member inline this.For (p: #seq<'T>, rest: 'T->'``MonadPlus<'U>``) : '``MonadPlus<'U>`` =
             let mutable isReallyDelayed = true
-            Delay.Invoke (fun () -> isReallyDelayed <- false; Unchecked.defaultof<'``MonadPlus<'U>``>) |> ignore
+            Delay.Invoke (fun () -> isReallyDelayed <- false; Empty.Invoke () : '``MonadPlus<'U>``) |> ignore
```
wallymathieu: :+1:

gusty: Note that there are no additional static members AFAIK.
Empty should be already there as it's a monadplus builder.

In the other case, return should be also there as it is a monad.

PR #277, opened 2020-02-20
Build on all fsproject branches

wallymathieu:
This means that for for some pull requests we get double builds

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,3 @@ script:
   - dotnet test -c Release tests/FSharpPlus.Tests
 env:
   FSHARP_FORMATTING_LOG=NONE
-branches:
-  only: 
-    - master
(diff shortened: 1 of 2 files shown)


adz:
Btw still doesn't build docs for me on Linux with dotnetcore only due to net47 ref... Will try on Windows later today

adz:
On windows I installed dotnetcore 3.1 and framework 4.8.
When i run "dotnet fsi buildDocs.fsx" it fails in similar way:
https://gist.github.com/adz/57ef4311771121c7589c5091eef0b5f0

wallymathieu:
@adz I have a PR where I've started to convert from paket-implicit dependency #275

adz:
Cool! I'll give it a try...

PR #278, opened 2020-02-21
Adz document generic ops

adz:
Addressed questions, https://github.com/fsprojects/FSharpPlus/pull/278/commits/ecd430114b6bd88d949684fbd3da12526ae24baf

...and then added these new commits:

Add two missed docs on choose and replace in Operators:
https://github.com/fsprojects/FSharpPlus/pull/278/commits/8111022e0fa3e146287c10979154761f0c5fd021

Add new index for Operators - first section is on "common" ones since they don't fall under any specific abstraction:
https://github.com/fsprojects/FSharpPlus/pull/278/commits/63d7b9f8393f6aae0125fe78e3035d72a0831043

I was thinking to detail the other generic operators in the appropriate abstraction page which often already does in it's 'minimal definition' and 'additional operator' sections. 
eg: http://fsprojects.github.io/FSharpPlus/abstraction-functor.html#map

If ok, I'll merge and squash before doing any more.

### DIFF
--- a/docsrc/content/generic-doc.fsx
+++ b/docsrc/content/generic-doc.fsx
@@ -22,7 +22,7 @@ Read about the specific operators:
 
  * Docs on [Operators - Common Combinators](operators-common.html)
  * Other docs exist for each [abstraction](abstractions.html)
- * API Doc for [Generic functions and operators](operators.html)
+ * API Doc for [Generic functions and operators](reference/operators.html)
 
 They're particularly useful in that the specific function called will
 depend on the input arguments and return type. However, this means you
@@ -78,7 +78,7 @@ implementation is used otherwise.
 
 What does this all mean?
 
-It means care is taken to use the most optimised impelemntation, and you can
+It means care is taken to use the most optimised implementation, and you can
 implement your own instances of generic functions if you implement the required
 methods.
 
--- a/docsrc/content/generic-doc.fsx
+++ b/docsrc/content/generic-doc.fsx
@@ -24,7 +24,45 @@ Read about the specific operators:
  * Other docs exist for each [abstraction](abstractions.html)
  * API Doc for [Generic functions and operators](operators.html)
 
+They're particularly useful in that the specific function called will
+depend on the input arguments and return type. However, this means you
+sometimes need to explicitly specify the type if this information is
+not available (actually it's a good debug technique to temporarily add
+the types explicitly when the compiler tells you that the types are wrong).
 
+For example:
+*)
+
+// Convert the number 42 to bytes... 
+// ... here the type is known (42 is an int, the return value is byte[])
+> let a = 42 |> toBytes;;  
+val a : byte [] = [|42uy; 0uy; 0uy; 0uy|]
+
+// However, this can't compile since the return type is not inferrable
+> let b = [|42uy; 0uy; 0uy; 0uy|] |> ofBytes;;  
+
+// The error will be something like:
+// 
+//  let b = [|42uy; 0uy; 0uy; 0uy|] |> ofBytes;;
+//  -----------------------------------^^^^^^^
+//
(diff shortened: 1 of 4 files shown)


docsrc/content/generic-doc.fsx
```diff
@@ -10,24 +10,80 @@ open FSharpPlus
...
 
- 3) Further user defined types
+An Invokable is written such that the most specific, and hence, optimised
+overload is resolved for existing .Net and F# types, and that a more general
+implementation is used otherwise.
 
- 4) Types defined in other libraries
+What does this all mean?
 
+It means care is taken to use the most optimised impelemntation, and you can
```
gusty: typo: implementation

src/FSharpPlus/Operators.fs
```diff
@@ -109,6 +116,7 @@ module Operators =
     /// Sequences two applicatives left-to-right, discarding the value of the second argument.
     let inline (<*  ) (x: '``Applicative<'U>``) (y: '``Applicative<'T>``): '``Applicative<'U>`` = ((fun (k: 'U) (_: 'T) -> k ) <!> x : '``Applicative<'T->'U>``) <*> y
 
+    /// Apply a lifted argument to a lifted function (flipped): arg <**> f
     let inline (<**>) (x: '``Applicative<'T>``) : '``Applicative<'T -> 'U>``->'``Applicative<'U>`` = flip (<*>) x
```
gusty: I don't find use cases for this operator, I wonder should we deprecate it and free it for other stuff?

src/FSharpPlus/Operators.fs
```diff
@@ -757,6 +778,13 @@ module Operators =
     /// <exception cref="System.ArgumentNullException">Thrown when the input collection is null.</exception>
     let inline sortByDescending (projection: 'T->'Key when 'Key : comparison) (source: '``Collection<'T>``) : '``Collection<'T>`` = SortByDescending.Invoke projection source
 
+    /// Splits a given ordered collection at each of the given sub-ordered collections
+    ///
+    ///     > "asdf" |> split ["s"];;
+    ///        val it : string list = ["a"; "df"]
+    ///
+    ///     > [1;2;3;4;5;6] |> split [ [2]; [5] ];;
+    ///        val it : int list list = [[1]; [3; 4]; [6]]
```
gusty: I'm not 100% sure including sample code in XML docs is common practice.
Though the sample is very illustrative.

wallymathieu: there are tags for code when documenting c#, these should be available in f# as well : 

https://docs.microsoft.com/en-us/dotnet/csharp/codedoc#example

src/FSharpPlus/Operators.fs
wallymathieu: yes, sounds reasonable

src/FSharpPlus/Operators.fs
adz: Thankyou for the reference! I thought it was all FSharp.Formatting that was reading and implementing these, and even wondered why you used xml, when that is supported markdown too. 
(this is my newness to .net showing).

I will read that link about codedoc and use these in future.

src/FSharpPlus/Operators.fs
adz: https://stackoverflow.com/questions/32426688/are-there-flipped-operators-for-applicative

> Control.Applicative has (<**>), which takes the arguments in the opposite order but is not flip (<*>) because effects are still sequenced from left to right.

Hmm, our implementation **is** flipped.

gusty: Then it's another reason to deprecate it, at least until v2.0 where we can eventually fix it.

wallymathieu:
I like the description of how f#+ works. I think it makes the library more approachable

PR #281, opened 2020-02-25
+Choice Invoker

gusty:
This PR will add an Invoker for the `choice` function.

It will allow to run the function over semigroups of Alt's (Alternatives without zero), which is a common use case.

Most overloads will use the technique of a left-zero knowledge to support short-circuit.

Note that Left Zero here (in contrast to the applicative implementation) is implemented in such a way that it's always defined (defaulting to false). Which simplifies a bit the constraint solving.

This also adds the `Reducible` abstraction which is a Foldable special case where there's no need for a Zero value, or initial state to fold. This means that the `reduce` function is safe.


tests/FSharpPlus.Tests/General.fs
```diff
@@ -1219,6 +1223,70 @@ module Alternative =
         Assert.AreEqual (SideEffects.get (), ["hello"])
         Assert.AreEqual (z, Some 1)
 
+    [<Test>]
+    let testChoice () =
```
wallymathieu: Great stuff 😄

PR #282, opened 2020-02-27
Format README.md header and add F# and Licence badges

ZaymonFC:


adz:
Thanks for the contribution @ZaymonFC !

I think it looks good -- but since the rest of the file is in normal markdown syntax, could you change it to suit?

In Markdown, just do an underline with "=" to get similar to <h1>.

Also, I'm in Adelaide, so I'm really stoked to see another Australian join in! 
Come say hello in the slack or gitter channel. :wave:

ZaymonFC:
@adz I've converted all the badges and associated links to the markdown syntax. Although I've left the header block wrapped in a `<center></center>`. Let me know if you want me to get rid of that too although I am a big fan of the centered header.

adz:
Thanks @ZaymonFC !

I liked it centered too, but not too fussed either way. Not sure what the markdown is for this, or if it requires HTML mixed in?

Please "Squash and merge" when you're ready.

Thanks for your contribution.

ZaymonFC:
@adz No worries. I don't think I can merge until you approve it since it's currently on `Changes requested`

adz:
Oh - right. Thanks @gusty .

PR #283, opened 2020-03-03
Restrict Fable compatibity

gusty:
Currently the project doesn't compile for Fable.

The main problem are the complex trait calls, those ones involving dummy parameters to control the overload resolution, Fable doesn't resolve them.

They have a double requirement: they have to pass the F# compiler, then Fable has to be able to resolve them. 

So, Fable expects to see a simple set of overloads (no DefaultX stuff) but F# doesn’t like it (fails with ambiguity or default to one of them).

This PR restricts Fable compatibility back to Extensions, plus stuff from FSharpPlus.Data and simple trait calls like `>>=`.

It would also add Fable tests and CI (thanks to @wallymathieu ) to make sure from now we don't break anything in Fable and that we can keep  supporting at least our Fable test code.

### DIFF
--- a/tests/FSharpPlusFable.Tests/FSharpTests/Extensions.fs
+++ b/tests/FSharpPlusFable.Tests/FSharpTests/Extensions.fs
@@ -4,6 +4,7 @@ open Testing
 open FSharpPlus
 open System.Collections.Generic
 
+open FSharpPlus.Data
 
 
 let ExtensionsTest = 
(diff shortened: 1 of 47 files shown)


tests/FSharpPlusFable.Tests/FSharpTests/Extensions.fs
```diff
@@ -36,4 +44,41 @@ let ExtensionsTest =
                    let r1 = m1 |> Map.union m2
                    let r2 = m1 |> Map.unionWith konst m2
                    equalMap r1 r2)
-  ]
\ No newline at end of file
+
+
+      testCase "Bind"
```
wallymathieu: 👍

PR #295, opened 2020-03-16
Link to type docs

wallymathieu:
In order to make it easier to find the type documentation from the reference documentation


wallymathieu:
A better solution would perhaps be to make sure that the links are absolute in order for the links to work in all xml-doc clients that understand markdown links.

wallymathieu:
I'm closing this since it's not appropriate to mix xml doc and reference docs.

PR #305, opened 2020-03-29
+NonEmptyList functions

wallymathieu:


src/FSharpPlus/Data/NonEmptyList.fs
```diff
@@ -39,6 +39,11 @@ module NonEmptyList =
     let singleton x = {Head = x; Tail = []}
     let toList {Head = x; Tail = xs} = x::xs
     let toSeq  {Head = x; Tail = xs} = seq { yield x; yield! xs; }
+    let toArray nel = toList nel |> List.toArray
+    let ofArray (array : _ array) = match Array.tryHead array with | Some head -> create head (Array.tail array |> Array.toList) | _ -> failwith "cannot construct empty non empty list"
```
gusty: Isn't it better to use `invalidOp` instead of `failwith` ?

src/FSharpPlus/Data/NonEmptyList.fs
```diff
@@ -39,6 +39,11 @@ module NonEmptyList =
...
+    let ofList (list : _ list) = match List.tryHead list with | Some head -> create head (List.tail list) | _ -> failwith "cannot construct empty non empty list"
```
gusty: Can't we remove `Unchecked.toNonEmptyList` now?

src/FSharpPlus/Data/NonEmptyList.fs
```diff
@@ -39,6 +39,11 @@ module NonEmptyList =
...
+    let ofSeq (list : _ seq) = match Seq.tryHead list with | Some head -> create head (Seq.tail list |> Seq.toList) | _ -> failwith "cannot construct empty non empty list"
```
gusty: Can we add some XML comments, stating in the remarks tag that it throws?

src/FSharpPlus/Data/NonEmptyList.fs
wallymathieu: Sure :smil

src/FSharpPlus/Data/NonEmptyList.fs
wallymathieu: sounds reasonable

src/FSharpPlus/Data/NonEmptyList.fs
```diff
@@ -35,15 +35,32 @@ type NonEmptyList<'t> = {Head: 't; Tail: 't list} with
...
+    /// <returns>Non empty list containing the elements of the array.</returns>
+    /// <exception cref="System.ArgumentException">Thrown when the input array is empty.</exception>
+    /// <remarks>Throws exception for empty array</remarks>
+    let ofArray (array : _ array) = match Array.tryHead array with | Some head -> create head (Array.tail array |> Array.toList) | _ -> invalidArg "array" "The input array was empty."
+    /// <summary>Builds a non empty list from the given list.</summary>
+    /// <param name="list">The input list.</param>
+    /// <returns>Non empty list containing the elements of the list.</returns>
+    /// <exception cref="System.ArgumentException">Thrown when the input list is empty.</exception>
+    /// <remarks>Throws exception for empty list</remarks>
+    let ofList (list : _ list) = match List.tryHead list with | Some head -> create head (List.tail list) | _ -> invalidArg "list" "The input list was empty."
```
xepaul: perhaps the use of Option can be removed? I'm not sure it will be optimized away
let ofList = function x::xs -> create x xs | [] -> invalidArg "list" "The input list was empty."

src/FSharpPlus/Data/NonEmptyList.fs
```diff
@@ -35,15 +35,32 @@ type NonEmptyList<'t> = {Head: 't; Tail: 't list} with
...
     let toArray nel = toList nel |> List.toArray
-    let ofArray (array : _ array) = match Array.tryHead array with | Some head -> create head (Array.tail array |> Array.toList) | _ -> failwith "cannot construct empty non empty list"
-    let ofList (list : _ list) = match List.tryHead list with | Some head -> create head (List.tail list) | _ -> failwith "cannot construct empty non empty list"
-    let ofSeq (list : _ seq) = match Seq.tryHead list with | Some head -> create head (Seq.tail list |> Seq.toList) | _ -> failwith "cannot construct empty non empty list"
+    /// <summary>Builds a non empty list from the given array.</summary>
+    /// <param name="array">The input array.</param>
+    /// <returns>Non empty list containing the elements of the array.</returns>
+    /// <exception cref="System.ArgumentException">Thrown when the input array is empty.</exception>
+    /// <remarks>Throws exception for empty array</remarks>
+    let ofArray (array : _ array) = match Array.tryHead array with | Some head -> create head (Array.tail array |> Array.toList) | _ -> invalidArg "array" "The input array was empty."
```
xepaul: Can the use of tryHead be removed seems redundant as the same fail is raised in Array.head

src/FSharpPlus/Data/NonEmptyList.fs
gusty: I agree, it's better to pattern match directly on the list.
But let's not use `function` which would create an implicit parameter in user-facing function.

src/FSharpPlus/Data/NonEmptyList.fs
gusty: It might be better to use `Array.isEmpty` and `array.[0]`.

src/FSharpPlus/Data/NonEmptyList.fs
wallymathieu: I've updated the implementation, what's your thoughts on it?

Issue #301, opened 2020-03-20
Question: working with monad transformer stacks


NickDarvey:
I've got a delightful stack of monad transformers here:
```fsharp
type Operation<'TResult, 'TError>   = ReaderT<SqlConnection * SqlTransaction, ResultT<Async<Result<'TResult, 'TError list>>>>
```

and it works well working with it in computation expressions.

```fsharp
  let run
    (get        : 'TSubject -> Operation<'TState, 'TAppError>)
    (subj       : 'TSubject) : Operation<unit, 'TAppError> = monad {
    let! state = get subj |> map (fun x -> x) 
    return ()
  }
```

`x` here is `'TState` which is lovely that map lifts that lambda all the way. However how do I do maps somewhere within the stack? Like transform the error type, represented here by `transformToCommonErrorTypeN`:

```fsharp
  let run
    (get        : 'TSubject -> Operation<'TState, 'TAppError>)
    (set        : 'TState   -> Operation<'TState, 'TOtherError>)
    (subj       : 'TSubject) : Operation<unit, 'TError> = monad {
    let! state = get subj |> transformToCommonErrorType1
    let state = state // ... interpret new state
    do! set subj state |> transformToCommonErrorType2
    return ()
  }
```
Given a function
`'TAppError -> 'TError`,
how do I lift that to a 
`Operation<'TResult, 'TAppError> -> Operation<'TResult, 'TError>`

(i.e. `ReaderT<SqlConnection * SqlTransaction, ResultT<Async<Result<'TResult, 'TAppError list>>>> -> ReaderT<SqlConnection * SqlTransaction, ResultT<Async<Result<'TResult, 'TError list>>>>`.)

So far I have:
```fsharp
// AppError :: 'TAppError -> 'TError`
let! state = get subj |> (AppError |> List.map |> mapError |> Async.map |> ResultT.??? |> ReaderT.map)
```
but I'm not sure what I need at the `ResultT.???` bit...

Side note: [ResultT.map has an odd return type](https://github.com/fsprojects/FSharpPlus/blob/3e5be186d2fece1fa3282bc7490f7f1ff97632c3/src/FSharpPlus/Data/Error.fs#L55) to my eye. Is `ResultT<'``Monad<'Result<('T -> 'U),'E>>``>` correct, or a mistaken type parameter name?

wallymathieu:
The generic type variable name is intended to be documentation around the intent of the type variable. So the intention is to guide you on how to use the monad transformer.

[...]

PR #306, opened 2020-03-31
Adds an example for stacked monad transformers

NickDarvey:
This example is based on @gusty answer to my question in https://github.com/fsprojects/FSharpPlus/issues/301.

It builds on the previous example in the doc showing how you can add one more layer.

### DIFF
--- a/docsrc/content/abstraction-monad.fsx
+++ b/docsrc/content/abstraction-monad.fsx
@@ -225,7 +225,7 @@ module CombineWriterWithResult =
 // For example, below we are stacking them like:
 // type Example = ReaderT<DateTime, ResultT<Writer<string list, Result<string * string * string, string>>>>)
 
-// Catch and throw is generic over all monad transformers in F#+ so you can use catch works in this example
+// Catch and throw is generic over all monad transformers in F#+ so catch works in this example
 // because there is a Result in the stack. We use it here to consolidate Result's 'TError.
 
 module CombineReaderWithWriterWithResult =
@@ -253,7 +253,9 @@ module CombineReaderWithWriterWithResult =
 
         return (x, y, z) }
 
-    let (_, log) = ew |> ReaderT.run >> ResultT.run >> Writer.run <| DateTime.UtcNow
+    let run expr = ReaderT.run expr >> ResultT.run >> Writer.run
+
+    let (_, log) = run ew DateTime.UtcNow
 
 
 // Many popular F# libraries are in fact an instantiation of a specific monad combination.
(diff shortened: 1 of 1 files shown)


docsrc/content/abstraction-monad.fsx
```diff
@@ -219,6 +219,42 @@ module CombineWriterWithResult =
...
 
 
+// You can also stack monad transformers.
+
+// A monad transformer and a monad is itself a monad, so you can pass that into another monad transformer.
+// For example, below we are stacking them like:
+// type Example = ReaderT<DateTime, ResultT<Writer<string list, Result<string * string * string, string>>>>)
+
+// Catch and throw is generic over all monad transformers in F#+ so you can use catch works in this example
+// because there is a Result in the stack. We use it here to consolidate Result's 'TError.
```
mitimes-engineering: This sentence doesn't read quite right...

adz: Oops, I was logged in a work service account. Was me in above ! :)

docsrc/content/abstraction-monad.fsx
```diff
@@ -219,6 +219,42 @@ module CombineWriterWithResult =
...
+
+    let ew = monad {
+        let! w = eitherConv divide5By       6.0
+        let! x = eitherConv divide5By       3.0
+        let! y = eitherConv divide5By       0.0
+        let! z = eitherConv otherDivide5By  0.0 </catch/> (throw << (fun _ -> "Unknown error"))
+
+        return (x, y, z) }
+
+    let (_, log) = ew |> ReaderT.run >> ResultT.run >> Writer.run <| DateTime.UtcNow
```
adz: Not sure about using `ew` here in docs.
Is it meaning "either_and_writer" (and ignore the reader bit)?

Also, maybe this should be broken down a little to show the intermediary parts?
It's hard to decipher `|>` `>>` `>>` and `<|` in one line, even though it is quite neat.

adz:
Nice example of the stack there! Thanks.

docsrc/content/abstraction-monad.fsx
gusty: > Also, maybe this should be broken down a little to show the intermediary parts?
It's hard to decipher |> >> >> and <| in one line, even though it is quite neat.

Yes, you have to know all the precedence rules very well to understand it.
Better to use `|>` and `( .. )`.

NickDarvey: > Not sure about using ew here in docs. Is it meaning "either_and_writer" (and ignore the reader bit)?
I assumed so. I built upon the previous example and `ew` was used there.

PR #310, opened 2020-04-03
Adds <= and >= Math operators to applicative

MajorZiploc:


gusty:
Thanks !
Could you add some tests, at least basic ones?

MajorZiploc:
Yes, of course. I was having some trouble building the test suite. I will look into it tomorrow and and some tests.

gusty:
Great, let us know if you can't solve the problems with the tests.

tests/FSharpPlus.Tests/General.fs
```diff
@@ -1162,6 +1163,29 @@ module Applicative =
         Assert.AreEqual ([113; 213; 123; 223; 114; 214; 124; 224], testVal)
         Assert.IsInstanceOf<Option<list<int>>> (Some testVal)
 
+        let testLTE1 = (Some 1) .<=. (Some 2)
```
gusty, line 1166: Do we really need enclosing parens here?
I think the operator alone should be enough.

Issue #191, opened 2019-08-26
Add functions for Nullable<_>


gusty:
I think at least `Option.ofNullable`should be added.

@mausch did a [whole module](http://www.fssnip.net/1A/title/Functions-around-nullable-types) some years ago, I don't think we should add all of them. Remember that Nullable are restricted to non value types, so things like `Option.map` would break.

cmeeren:
`Option.ofNullable` has been part of FSharp.Core for some time now.

gusty:
Thanks, for some reason I didn't realize it was already there. So I would say let's have a look at the other functions and if there's nothing interesting to add, that justify that module, we can close this issue.

cmeeren:
There's `Option.toNullable` built-in.

wallymathieu:
Thanks @cmeeren

PR #313, opened 2020-04-06
Add Nullable module and generics map and bind

mattstermiller:


src/FSharpPlus/Control/Monad.fs
```diff
@@ -32,6 +32,7 @@ type Bind =
     static member        (>>=) (source             , f: 'T -> _           ) = async.Bind (source, f)                                        : Async<'U>
     static member        (>>=) (source             , k: 'T -> _           ) = Result.bind k source                                          : Result<'U,'E>
     static member        (>>=) (source             , k: 'T -> _           ) = Choice.bind k source                                          : Choice<'U,'E>
+    static member        (>>=) (source             , f: 'T -> _           ) = Nullable.bind   f source                                      : Nullable<'U>
```
wallymathieu, line 35: Perhaps remove these overloads in order for the tests to compile?

wallymathieu:
The overloading of map and bind are probably a bit problematic. What do you say about only adding the Nullable module in order to be able to iterate on the nullable overloads later on?

mattstermiller:
Those overloads were the whole reason for the PR. I wanted Nullable support for the generic map and bind operators. I can easily create a Nullable module in my own utility library.

I don't understand the type sorcery going on to even make an educated guess as to how to move forward with this. I can't even get the test project from the tip of master to compile on my machine with the latest dotnet SDK.

mattstermiller:
OK, I just found a comment in another issue stating that debug builds don't compile. It seems like there should be a huge warning somewhere about this. The fact that it doesn't compile is a huge turn-off to a new contributor.

wallymathieu:
What happens if you try to build it using `dotnet build -c Release` ? There is an f# bug that causes the compilation to fail for debug builds.

src/FSharpPlus/Control/Functor.fs
```diff
@@ -81,6 +81,7 @@ type Map =
     static member Map ((x: StringBuilder       , f        ), _mthd: Map) = new StringBuilder (String.map f (string x))
     static member Map ((x: Set<_>              , f        ), _mthd: Map) = Set.map f x
     static member Map ((_: Set2<'T>            , _: 'T->'U), _mthd: Map) = Set2<'U>()
+    static member Map ((x: Nullable<_>         , f        ), _mthd: Map) = Nullable.map f x
```
wallymathieu, line 84: The compilation issue is due to the fact that it chooses Nullable over other overloads when resolving the correct overload. If you look further down in the file you'll notice that other overloads have the priority set using `Default1` and `Default2`.

mattstermiller: Wow, thanks! That works on my machine.

wallymathieu:
The issue around the Fable compilation is due to the fact that nullable does not exist in Fable, why you need to surround nullable module it by an `#if` block.

tests/FSharpPlus.Tests/Extensions.fs
```diff
@@ -267,4 +268,18 @@ module Extensions =
 
   [<Test>]
   let ``Array.lift2 should combine all arrays `` () =
-    areStEqual [|11; 21; 31; 12; 22; 32|] (Array.lift2 (+) [|1;2|] [|10;20;30|])
\ No newline at end of file
+    areStEqual [|11; 21; 31; 12; 22; 32|] (Array.lift2 (+) [|1;2|] [|10;20;30|])
+
+#if !FABLE_COMPILER
```
wallymathieu, line 273: This test project is not compiled for Fable unfortunately.

gusty:
What's the status on this?
Is it ready for review?
If the idea is to keep the generic overloads I would add some tests for them.

gusty:
I see there are currently some tests, included in the same test for the modules, that's fine but they seems to be monomorphic (int to int). Can you add polymorphic ones? Like mapping/binding from int to other type?

mattstermiller:
I've added polymorphic functions to the map and bind tests and added a test for `iter`.

If you would prefer me to move the generic function tests, please let me know where they should go.

Thank you for your interest in this request :)

gusty:
That makes sense. I'll review it and most likely I will accept it.
The only part that I'm a bit unsure is whether to include the generic versions in v1.1 because we already did 4 RCs and I don't want to risk adding something new just for the release.
But will see what can I do.

Thanks for contributing back !

mattstermiller:
@gusty I have another commit that adds more functions to this Nullable module for #191. Would you like me to do a separate PR for that, or just push another commit to this one?

gusty:
Great, do it here.
Later we can always either rename this PR, or separate in a second one.

Issue #307, opened 2020-04-03
RC4 - DList isn't fully Fable compatible


Shmew:
Fable isn't able to type test generic arguments, so [this line](https://github.com/fsprojects/FSharpPlus/blob/master/src/FSharpPlus/Data/DList.fs#L40) when compiled via Fable gives a warning that it isn't valid.

gusty:
Yes, I saw that.

But I am not sure how typically you define equality in Fable. Do you know how?

Shmew:
I've always just worked around it in the past, I played around in the repl for a while with no success. I'm not sure if it's possible for this type of scenario?

Maybe @MangelMaxime or @ncave could chime in about it.

ncave:
@Shmew See if anything here helps: https://github.com/fable-compiler/Fable/issues/1981

Shmew:
@gusty did ncave's link help?

gusty:
Looks helpful, tbh I still have to study all the possibilities.
Do you have a specific suggestion?

Shmew:
I have no idea which would be optimal to be honest, that's why I figured I'd leave the implementation to you 😅

wallymathieu:
Well, perhaps we could pair program it?

gusty:
@Shmew I read all the options but none of them seem to address our issue.

I did a forced cast for Fable, which in theory would throw exceptions but AFAIK the type checking process would prevent from comparing 2 different types.

Have a look at the PR #314

Shmew:
Sorry I haven't been able to reply sooner! I've been pretty busy. It looks fine to me, if the tests pass then 👍.

PR #314, opened 2020-04-12
+ Fix DList equality for Fable

gusty:
Address #307 

Add tests to pin and document behavior.

### DIFF
--- a/tests/FSharpPlusFable.Tests/FSharpTests/Extensions.fs
+++ b/tests/FSharpPlusFable.Tests/FSharpTests/Extensions.fs
@@ -87,4 +87,14 @@ let ExtensionsTest =
                    equal (mapItem2 string (1,2,3)) (1,"2",3)
                    equal (item3 (1,2,3)) 3
                    )
+
+      testCase "eq on DList"
+          (fun () ->
+                     let a = DList.ofSeq [1;2;3]
+                     let b = DList.ofSeq [1;2;3]
+                     let c = DList.ofSeq [1;2]
+                     equal true (a = b)
+                     equal false (a = c)
+                     )
+
 ]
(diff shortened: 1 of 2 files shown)


gusty:
Note that the last test differs in Fable.
But that's also true for normal lists, I think there's no way around it since JS doesn't carry type information.

tests/FSharpPlusFable.Tests/FSharpTests/Extensions.fs
```diff
@@ -87,4 +93,11 @@ let ExtensionsTest =
                    equal (mapItem2 string (1,2,3)) (1,"2",3)
                    equal (item3 (1,2,3)) 3
                    )
+
+      testCase "eq on DList 1" (fun () -> equal true  (dlistA = dlistB))
+      testCase "eq on DList 2" (fun () -> equal false (dlistA = dlistC))
+      testCase "eq on DList 3" (fun () -> equal true  ((dlistA :> obj) = (dlistB :> obj)))
+      testCase "eq on DList 4" (fun () -> equal false ((dlistA :> obj) = (dlistC :> obj)))
+      testCase "eq on DList 5" (fun () -> equal true  ((dlistA :> obj) = (dlistD :> obj))) // this behavior differs from (non-fable) F# but same way it would be with normal lists.
```
gusty, line 101: Note that the last test differs in Fable.
But that's also true for normal lists, I think there's no way around it since JS doesn't carry type information.

wallymathieu: Unless you add additional encoding using something like ```{"$type":"obj", "value":1}``` in order to include the additional information at runtime.

Issue #288, opened 2020-03-07
Roadmap V2.0


gusty:
I think for V2.0 we can start discussing features here, before moving to a formal project.

What comes to my mind at the moment is:

1. Remove subsumption from Applicatives: Right now `seq` "eats" all custom types implementing `seq`

2. Same for Traversable, same situation: although this might not require a breaking change (todo find out)

3. Implement IEnumerable for Vectors, now that (1) has to be done.

4. New `ListT` monad transformer, possibly new `SeqT` as well, coded using a similar technique as the one used for Free Monads.

5. Review computation expressions: autosense will be removed but we have to do something to make it easy for users to understand why a strict monad can't be used with the non-strict CE. Maybe make compilation fail when calling a TryWith?

6. Add ValueTuple, ValueTask and other latest addition to the framework. We can require a higher framework version

7. Consider suggestions by @dsyme regarding using the return type for type inference.

Am I missing something?

adz:
I would like to organise namespace so that we: 1) can more easily generate docs per functional area (thinking of how operations is separated nicely into sections by banner comment, but API docs are resorted alphabetically and all at same level), 2)  reduce possible instances by opening only what you're using.

The later requires to split the Invokables so that each implentation sits with the 'openable' type. Such as Arrow, or NonEmptyList, etc.

This should help considerably with compiler error messages as there is reduced implementations. I also wonder if it would help speed up compiler time?

I'm not exactly sure on how to arrange for optimal dev work flow, but could be that there is separation between common and exotic abstractions, or maybe that there is 'open common' and otherwise opt in for each type... 

WDYT?

gusty:
No doubt namespaces should be re-organized.

I personally don't see a big benefit on splitting Invokables in different namespaces, but at the same time I don't see any harm on doing that, since they are rarely used in client code directly.

Separation between common and exotic abstraction could be interesting but hard to draw the line.

One thing I definitely would like to change is a few extensions (like `List.traverse`) which lives in `.Data` right now. I think they don't belong there, since `List`  is not in `.Data`.

So I would say, if/when you have a concrete proposal in mind please drop the details and we can analyze it, because this is the right time to do it ;)

NinoFloris:
https://github.com/fsprojects/FSharpPlus/issues/245

I would vote for defaulting to left for success right for error

cmeeren:
You could consider removing functions that are part of FSharp.Core, e.g. List.singleton.

[...]

PR #315, opened 2020-04-25
Rename string trim functions

cmeeren:
As requested in https://github.com/fsprojects/FSharpPlus/issues/288#issuecomment-619355229


wallymathieu:
Looks good. Looks like `trimboth` is used in python. Another alternative is to use the word `strip`.

wallymathieu:
It would be good to hear some people chime in here 😃

cmeeren:
I think we should stick to some version of `trim` because it's what .NET users expect.

Alternatives to `trimBoth` are `trimStartEnd`, or `trimChars` (though then it's not consistent with `trimStart` and `trimEnd`).

wallymathieu:
That's true. Let's see what people says 😃

gusty:
I think I would be in favor of this change, but of course in case we approve it, it will have to be for v2.0

In the meantime it would be interesting to hear other voices. Maybe @cannorin can comment as he added many string functions to the lib.

PR #321, opened 2020-05-05
Fix intercalate's generic overload

### LINKED ISSUE #320: Regressions with generic intercalate function
I just realized the `intercalate` function doesn't fully work, specifically it fails when resolving to the more generic overload.

Here's a small repro: `intercalate 10 (seq [1;2;3])`

Looking back in time it seems that when it was introduced, this problem [was spotted](https://github.com/fsprojects/FSharpPlus/commit/7b0465568356f4b29e4fb6dc802d640721eb9694#diff-66e84162b5bc78265f8ed0ef828e14aeL310-R312) but the comment there makes me think I tried a preview of F#4.1 (as I believe F#4.1 wasn't ready at that time) and it kind of worked.

Now testing it back with 4.1 I see it doesn't work, but the funny thing is that I get a completely different error, here's a summary:

 - F# 4.1 -> Fails with an overload resolution error of the `Intercalate` method
 - F# 4.5 -> Fails with an overload resolution error of the `ToSeq` method
 - F# 4.7 -> Fails with an internal error: `Undefined or unsolved type variable`

I can workaround it now by commenting out the first default which is less generic and just an optimization, but it's interesting to note all these changes along all F# compiler version number.

Moreover I would like to test it with other compiler version, like F# 4.0 and with the future RFC for extensions visible which has a fix for some `Undefined or unsolved type variable` errors.

For now I would create a fix with the above mentioned workaround and as I collect more information I will add to that list.

If someone is willing to test further, the above list is done testing against the version 1.1 of this library.

gusty:
This will fix #320 for F# versions 4.7 and lower by removing the Default1 overload which is an optimization.

It is expected that future versions of the F# compiler will fix this issue, so this patch won't be needed anymore.

Therefore in the future event of a release of this library which targets a min compiler version which doesn't have this issue, this patch can be revisited and removed.

### DIFF
--- a/src/FSharpPlus/Control/Collection.fs
+++ b/src/FSharpPlus/Control/Collection.fs
@@ -421,8 +421,7 @@ type Intersperse =
 type Intercalate =
     inherit Default1
 
-    static member inline Intercalate (x: '``Foldable<'Monoid>``, e: 'Monoid          , [<Optional>]_impl: Default2   ) = let f t x = match t, x with (true, _), x -> false, x | (_, acc ), x -> (false, Plus.Invoke (Plus.Invoke acc e) x) in Fold.Invoke f (true, Zero.Invoke ()) x |> snd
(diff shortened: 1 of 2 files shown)


wallymathieu:
Could someone have used the overload?

gusty:
Yes, it's a real possibility that someone used already the overload I removed. If that's the case now it will resolve to the more generic one, which will yield the same result maybe in a less efficient way.

PR #330, opened 2020-07-01
Move include closer to usage

wallymathieu:


wallymathieu:
Yes, we can do that. I'll have a look at other places where it can make sense.

PR #347, opened 2020-08-30
Fix try blocks for lazy / strict monads

### LINKED ISSUE #137: Weird try/with/finally behavior in monad CE
It doesn't seem that `try` catches exceptions. Repro (using xUnit and Unquote):

```f#
/// Helper type for simulating side effects.
type Trigger() =
  let mutable isTriggered = false
  member __.Trigger () = isTriggered <- true
  member __.Triggered = isTriggered

[<Fact>]
let ``behavior of try-with-finally when thrown`` () =
  let tTry = Trigger()
  let tCatch = Trigger()
  let tFinally = Trigger()

  let result = monad {
    try
      try
        failwith ""
        tTry.Trigger()
      with _ -> tCatch.Trigger()
    finally tFinally.Trigger()
  }
  test <@ result = Ok () @>
  test <@ not tTry.Triggered @>
  test <@ tCatch.Triggered @>
  test <@ tFinally.Triggered @>
```

None of the unquote tests are checked, because the exception is not caught, but they do describe what should happen if the exception is caught.

### LINKED ISSUE #346: Monad computation does not work with TryWith as expected when used with Result
### Description

My logic here is that computation expressions should not deviate from universal concepts as that would surprise and confuse users. A `try...with` operation should always handle exceptions and not propagate them (in other words it should behave similarly to the `try...with` expression outside of a computation expression).

The `monad.strict` works as expected in this case, but `monad` does not.

### Repro steps

```fsharp

let r : Result<unit, Exception> = monad {
    try
        failwith "oops"
        return ()
    with exn -> return! Error exn
}

```

### Expected behavior

Error

### Actual behavior

Exception

gusty:
This will generate a compile error when attempting to use a try .. with construction inside a non-strict computation expression over a type that doesn't provide a `TryWith` implementation.

Same applies to `TryFinally`.

For reference: http://tryjoinads.org/docs/computations/monads.html

 _For monadic computations that encapsulate untracked F# effects, the members TryWith and TryFinally cannot be implemented in terms of other operations. They need to understand the structure of the computation, in order to wrap the appropriate call that actually runs the effects with try .. with or try .. finally._ 

It also fixes the behavior of lazy monad transformers by adding a special Invoker for them, which takes strict monad defaults into consideration, allowing the composition of lazy with strict monads, only in the case where all monads in the stack are strict, user will get the above compiler error. This is the correct behavior, since a monadic composition is strict only when all composed monads are strict.

This closes #346 and closes #137

- [x] Apply compiler error when the monad is not strict
- [x] Add another Invoker to be used within Monad Transformers
- [x] Upgrade State and Reader, add tests
- [x] See if we need to upgrade Cont, add tests
- [x] Do the same for `TryFinally`, check that Using doesn't break

### DIFF
--- a/src/FSharpPlus/Control/Monad.fs
+++ b/src/FSharpPlus/Control/Monad.fs
@@ -175,6 +175,23 @@ type TryWith =
         let inline call (mthd: 'M, input: 'I, _output: 'R, h: exn -> 'I) = ((^M or ^I) : (static member TryWith : _*_*_ -> _) input, h, mthd)
         call (Unchecked.defaultof<TryWith>, source, Unchecked.defaultof<'``Monad<'T>``>, f)
 
+type TryWithStrict =
+    inherit Default1
+
+    static member        TryWith (computation: unit -> '``Monad<'T>``, catchHandler: exn -> '``Monad<'T>``, _: Default3) = try computation () with e -> catchHandler e
+    static member inline TryWith (computation: unit -> '``Monad<'T>``, catchHandler: exn -> '``Monad<'T>``, _: Default1) = (^``Monad<'T>`` : (static member TryWith : _*_->_) computation (), catchHandler) : '``Monad<'T>``
+    static member inline TryWith (_: unit -> ^t when ^t: null and ^t: struct, _    : exn -> 't            , _: Default1) = ()
+    
+    static member        TryWith (computation: unit -> seq<_>        , catchHandler: exn -> seq<_>       , _: Default2     ) = seq (try (Seq.toArray (computation ())) with e -> Seq.toArray (catchHandler e))
+    static member        TryWith (computation: unit -> 'R -> _       , catchHandler: exn -> 'R -> _      , _: Default2     ) = (fun s -> try (computation ()) s with e -> catchHandler e s) : 'R ->_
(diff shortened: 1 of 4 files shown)


wallymathieu (approved):
I've tried to compile the following code and get the expected compilation error: 
```f#
let r : Option<Exception> = monad {
    try
        failwith "oops"
        return! None
    with exn -> return! Some exn
}
```

gusty:
We need to test how this will play with Monad Transformers.

There, the TryWith implementations call the TryWith of the other monad, which could be strict.

This could be a problem, but it's possible that this problem is present already, I mean before this PR or not. We need to find out.

wallymathieu:
So perhaps we should add some more tests of monad transformers then.

gusty:
Yes, actually this PR should have started with those tests, so we can compare the bahaviour before and after.

Test should look like this:

```
let monadTransformerTest3 () =
    let x : StateT<unit, list<unit*unit>> = monad {
        try
            failwith "uncaught exception"
        with _ ->
            printfn "exception handled" // ideally this will be a sideEffect
            Unchecked.defaultof<_>
    }
    x

    (monadTransformerTest3 () |> StateT.run ) ()
    // check the side effect executed

```

and the same with other monad combinations, here's a Lazy with a strict, we can add the remaining 3 cases.

wallymathieu:
Seems like it doesn't compile on the branch and fails on master.

PR #352, opened 2020-09-07
Fix bug: wrong parameters order for externals

gusty:
In collections, some parameter order from the Invoker on instance (external types) is wrong.


wallymathieu:
Should we add this to v2 since it's a breaking change?

gusty:
Well, I would consider it a bug.
But we need to fix it in 1.1 as well.

wallymathieu:
So since no one seems to have noticed it, they have probably not used the wrong definition?

gusty:
Impossible. This is safe to merge and include in 1.x releases.

PR #355, opened 2020-09-14
More docs - starting to go through the extensions

adz:
Have a bit more time to get back into FSharpPlus so started in on some more API docs...

Now documents over Extensions pretty completely. Tried to follow conventions.

Please review the "Enumerator" file as it was a little tricky, choosing what to omit, etc.

### DIFF
--- a/src/FSharpPlus/Extensions/Array.fs
+++ b/src/FSharpPlus/Extensions/Array.fs
@@ -6,7 +6,7 @@ module Array =
 
     open System
 
-    /// <summary>Applies an array of functions to an array of values and concatenates them</summary>
+    /// <summary>Applies an array of functions to an array of values and concatenates them.</summary>
     /// <param name="f">The array of functions.</param>
     /// <param name="x">The array of values.</param>
     /// <returns>A concatenated array of the resulting arrays from applying each function to each value</returns>
@@ -35,7 +35,7 @@ module Array =
     /// Creates a sequence of arrays by splitting the source array on any of the given separators.
     let split (separators: seq<_ []>) (source: _ []) = source |> Array.toSeq |> Seq.split separators |> Seq.map Seq.toArray
 
-    /// Replace a subsequence of the source array with the given replacement array.
+    /// Replaces a subsequence of the source array with the given replacement array.
     let replace (oldValue: _ []) (newValue: _ []) source = source |> Array.toSeq |> Seq.replace oldValue newValue |> Seq.toArray : 'T []
 
     /// <summary>
@@ -77,7 +77,7 @@ module Array =
 
     /// <summary>
     /// Creates two arrays by applying the mapper function to each element in the array
-    /// and classifying the transformed values depending on whether they were wrapped with Choice1Of2 or Choice2Of2.
+    /// and classifies the transformed values depending on whether they were wrapped with Choice1Of2 or Choice2Of2.
     /// </summary>
     /// <returns>
     /// A tuple with both resulting arrays.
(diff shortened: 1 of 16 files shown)


gusty:
I think it is because of the markdown.
I will have a look and try to identify which part breaks FSharp Formatting.
It happened to me before.

src/FSharpPlus/Extensions/Seq.fs
```diff
@@ -20,6 +20,17 @@ module Seq =
...
 
+    /// <summary>Applies a sequence of functions to a sequence of values and concatenates them</summary>
+    /// <param name="f">The seq of functions.</param>
+    /// <param name="x">The seq of values.</param>
+    /// <returns>A seq concatenating the results from applying each function to each value</returns>
+    /// 
+    /// <example>
+    /// <code>
+    /// > Seq.apply [double; triple] [1; 2; 3];;  
+    /// val it : seq<int> = seq [2; 4; 6; 3; ...]
```
wallymathieu, line 31: You probably need to xml-escape the `<int>`

gusty: Yes, it was that. The generic brackets. At least in my case.

adz: Thanks!

src/FSharpPlus/Extensions/Array.fs
```diff
@@ -6,6 +6,17 @@ module Array =
 
     open System
 
+    /// <summary>Applies an array of functions to an array of values and concatenates them</summary>
```
gusty, line 9: Nitpicking: All sentences should end with a `.`

gusty (changes requested):
Nice job, I just commented about conventions, if you think you can align them a bit.

src/FSharpPlus/Extensions/Map.fs
```diff
@@ -9,12 +9,27 @@ module Map =
     open System.Linq
     #endif
 
+    /// NOTE: Missing from other dictionary like modules:
```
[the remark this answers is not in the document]
adz: Yes indeed! Was a note to myself - I considered adding them right here, then decided to do it in separate PR.

src/FSharpPlus/Extensions/Dict.fs
```diff
@@ -8,18 +8,49 @@ module Dict =
 
     #if !FABLE_COMPILER
     open System.Linq
+
+    /// Convert an IDictionary to an IReadOnlyDictionary
```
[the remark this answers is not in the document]
adz: Pretty sure it wasn't me, but I can see the inconsistency now. :+1:

src/FSharpPlus/Extensions/Array.fs
adz: Have done, and also reviewed all doc strings in Extensions.

Note: I generally treat doc summaries as a title without a full stop, however even I was inconsistent as I copied some existing comments sometimes!

src/FSharpPlus/Extensions/Seq.fs
```diff
@@ -37,17 +36,24 @@ module Seq =
...
-    /// keys tupled with values. Each key contains an array of all adjacent elements that match 
-    /// to this key, therefore keys are not unique but they can't be adjacent
-    /// as each time the key changes, a new group is yield.</summary>
+    /// <summary>
+    /// Chunks the seq up into groups with the same projected key by applying
+    /// the key-generating projection function to each element and yielding a sequence of 
+    /// keys tupled with values.
+    /// </summary>
+    ///
+    /// <remarsk>
```
gusty, line 45: This typo is making the build to fail.

gusty (changes requested):
Looks good. Existing docs are not always standard, but we can align them as we go.

gusty (approved):
Excellent work !

PR #357, opened 2020-09-19
Add NonEmptySeq

### LINKED ISSUE #354: NonEmptySeq
Given that we're going to add some NonEmpty collections, it will be natural to wish we had a NonEmptySeq that could be used as source for creating / transforming between them.

I think we can ensure non-emptiness  by using specific constructors:

 - We can create from a single value, alternatively taking an additional normal `seq` as "tail".
 - We can define a closed range operator that return a `NonEmptySeq` `|..`, also `..|` and `|..|` where the number besides the bar always exists in the range. Example `1|..3 -> NonEmptySeq [1;2;3]`, `1|..0 -> NonEmptySeq [1]`, `10|..|3 -> NonEmptySeq [3]`.
 - As in NonEmptyList, we can define a builder that doesn't allow the Zero method, for NonEmptySeq literals: `nes {1;2;3}`
 - Finally, we can add some safe/unsafe constructors from regular sequences, by checking on `Seq.isEmpty`

Any thoughs?

cannorin:
Closes #354.

TODO:

- [x] Implement NonEmptySeq
- [x] Add tests

### DIFF
--- a/src/FSharpPlus/Control/Alternative.fs
+++ b/src/FSharpPlus/Control/Alternative.fs
@@ -4,6 +4,7 @@ namespace FSharpPlus.Control
 
 open System.Runtime.InteropServices
 open FSharpPlus
+open FSharpPlus.Data
 open FSharpPlus.Internals
 
 
@@ -28,6 +29,7 @@ type Empty =
 type Append =
     inherit Default1
     static member        ``<|>`` (x: 'T seq              , y              , [<Optional>]_mthd: Default2) = Seq.append   x y
+    static member        ``<|>`` (x: 'T NonEmptySeq      , y              , [<Optional>]_mthd: Default2) = NonEmptySeq.append x y
 
     static member inline ``<|>`` (x: '``Alt<'T>``        , y: '``Alt<'T>``, [<Optional>]_mthd: Default1) = (^``Alt<'T>`` :  (static member (<|>) : _*_ -> _) x, y) : '``Alt<'T>``
     static member inline ``<|>`` (_: ^t when ^t: null and ^t: struct   , _,             _mthd: Default1) = ()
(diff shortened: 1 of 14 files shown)


tests/FSharpPlus.Tests/ComputationExpressions.fs
```diff
@@ -64,7 +64,8 @@ module ComputationExpressions =
                     SideEffects.add "execute this"
             with
             | e -> 
-                SideEffects.add (sprintf "Exception! %s" e.Message)
+                // SideEffects.add (sprintf "Exception! %s" e.Message) // this can be different if run on non-English environment
+                SideEffects.add "Exception! Attempted to divide by zero."
```
wallymathieu, line 68: You could set the culture to be invariant culture when running the code, I think there might be attributes that you can paint the testfixture or tests.

wallymathieu:
```f#
open NUnit.Framework

[<Test;SetCulture("sv-SE")>]
let TestSwedish () =
    Assert.AreEqual("2,9", 2.9.ToString())
[<Test;SetCulture("ja-JP")>]
let TestJapanese () =
    Assert.AreEqual("2.9", 2.9.ToString())
```

wallymathieu: Though that doesn't seem to change how exceptions are localized, so perhaps better to catch more specific exception `DivideByZeroException` ?

src/FSharpPlus/Control/Alternative.fs
```diff
@@ -4,6 +4,7 @@ namespace FSharpPlus.Control
 
 open System.Runtime.InteropServices
 open FSharpPlus
+open FSharpPlus.Data
```
gusty, line 7: What about using a type alias `type NonEmptySeq<'t> = FSharpPlus.Data.NonEmptySeq<'t>` instead of opening the whole namespace? Just my 2 cents, feel free to leave it as it is.

gusty (approved):

src/FSharpPlus/Control/Alternative.fs
cannorin: I think `open FSharpPlus` should not add NonEmptySeq to the scope, so I'll keep this as is.

PR #365, opened 2020-10-07
Fix and enhance CE tests

gusty:
Q: Type was wrong, I wonder how test were passing, or would they break now :)

A: Yes, they break, but I can't read the expected result

So, we're using here areEquivalent to check side-effects and adding a helper to check directly side effects against expectations.

And of course, as last step we can fix our expectations and everything should be green again.

### DIFF
--- a/tests/FSharpPlus.Tests/ComputationExpressions.fs
+++ b/tests/FSharpPlus.Tests/ComputationExpressions.fs
@@ -209,7 +209,7 @@ module ComputationExpressions =
             let!  _  = j1 "" (Some len)
             return str }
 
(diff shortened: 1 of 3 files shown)


tests/FSharpPlus.Tests/ComputationExpressions.fs
```diff
@@ -19,20 +19,20 @@ module ComputationExpressions =
             SideEffects.add "2" }
 
         // Check side effects are not yet executed
-        areEqual [] (SideEffects.get ())
+        SideEffects.are []
```
wallymathieu, line 22: ❤️

PR #366, opened 2020-10-07
Add type alias for non-empty collections and builder for NonEmptySeq

### LINKED ISSUE #363: Builders / type abbreviations for NonEmpty collections
I propose we add a builder for `NonEmptySeq<_>` and type abbreviations for all 4 NonEmpty collections.

The question is, how do we call those builders / type abbreviations?

I can think of 2 options:
 

1. `neseq`, `nelist`, `nemap` and `neset`

2. `nseq`, `nlist`, `nmap` and `nset`

I think option 1 is being used in Haskell, with capital `NE`.

Anyway this is F# and although it's nice to align with other languages naming, we have different naming conventions.

Note: for the existing `nel` builder we can deprecate it, or just keep it as a second abbreviation, which won't be promoted.

Thoughts? Please, existing NonEmptyList users, chime in.

gusty:
This adds a type alias for all non empty collections and also a CE builder `NonEmptySeq` which will have the same identifier as its type alias.

The existing builder for `NonEmptyList` will match its type alias, aligning this way with `seq` (type alias and CE), the old `nel` remains available for backwards compatibility.

Finally this adds two range operators for non-empty sequences: `|..` and `..|`.

This will close #363

### DIFF
--- a/src/FSharpPlus/Data/NonEmptyList.fs
+++ b/src/FSharpPlus/Data/NonEmptyList.fs
@@ -33,6 +33,9 @@ type NonEmptyList<'t> = {Head: 't; Tail: 't list} with
     member this.Length = 1 + List.length this.Tail
 
 
+/// A type alias for NonEmptyList<'t>
+type nelist<'t> = NonEmptyList<'t>
+
 /// Basic operations on NonEmptyList
 [<RequireQualifiedAccess>]
 module NonEmptyList =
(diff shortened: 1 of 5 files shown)


src/FSharpPlus/Data/NonEmptyList.fs
```diff
@@ -262,6 +265,7 @@ module NonEmptyListBuilder =
         member __.Delay expr = expr ()
         member __.Run (x: NonEmptyList<_>) = x
     let nel = NelBuilder ()
```
wallymathieu, line 267: should this one be marked as obsolete?

wallymathieu (approved):

src/FSharpPlus/Data/NonEmptyList.fs
gusty: Yes, let's do that.

PR #368, opened 2020-10-09
Introduce choose function to Map, Dictionary, IDictionary and IReadOnlyDictionary

3Rafal:
Added in Extensions and in generic `choose`

### DIFF
--- a/src/FSharpPlus/Extensions/Dict.fs
+++ b/src/FSharpPlus/Extensions/Dict.fs
@@ -73,6 +73,19 @@ module Dict =
             | None    -> ()
         dct :> IDictionary<'Key, 'U>
 
+    /// <summary>Applies given function to each value of the given dictionary.</summary>
+    /// <param name="f">The mapping function.</param>
+    /// <param name="x">The input dictionary.</param>
+    ///
+    /// <returns>Returns Map with values x for each dictionary value where the function returns Some(x).</returns>
+    let choose f (x: IDictionary<'Key, 'T>) =
+        let dct = Dictionary<'Key, 'U> ()
+        for KeyValue(k, v) in x do
+            match f v with
+            | Some v -> dct.Add (k, v)
+            | None    -> ()
+        dct :> IDictionary<'Key, 'U>
+        
     /// <summary>Tuples values of two dictionaries.</summary>
     /// <remarks>Keys that are not present on both dictionaries are dropped.</remarks>
     /// <param name="x">The first input dictionary.</param>
--- a/src/FSharpPlus/Extensions/Dict.fs
+++ b/src/FSharpPlus/Extensions/Dict.fs
@@ -77,7 +77,7 @@ module Dict =
     /// <param name="f">The mapping function.</param>
     /// <param name="x">The input dictionary.</param>
     ///
-    /// <returns>Returns Map with values x for each dictionary value where the function returns Some(x).</returns>
+    /// <returns>Returns dictionary with values x for each dictionary value where the function returns Some(x).</returns>
     let choose f (x: IDictionary<'Key, 'T>) =
         let dct = Dictionary<'Key, 'U> ()
         for KeyValue(k, v) in x do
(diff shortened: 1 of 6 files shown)


gusty:
Nice. I would like to include it in the coming release.

Could you add some test for the generic `choose`?

The question still remains about how to call it, I mean the non generic version. I think `chooseValues` is not that nice but I fear one day FSharp.Core could include `choose` operating on keys and values and would cause a problem. 

@cartermp what do you think?

3Rafal:
> Could you add some test for the generic `choose`?

I've added tests similar to the tests of generic `map`.

cartermp:
Hard to say. We recently added `change` that changes a value stored under a single key, so something like `choose` could be done. But it might not use that same name.

gusty:
I'm thinking we can always change it to `choose`, and deprecate `chooseValues` but the other way around might be complicated.

Also while writing this, I realize that FSharp.Core `choose` name is a bit unfortunate, because it should be called `chooseBy` in order to align with the other functions, while `choose` alone should do only the flattening. This is the reason why we see a lot of `choose id` in F# codebases.

@adz @cannorin @wallymathieu what to you think?

3Rafal:
I see your point, and I agree that there is some risk of single-value `choose` being introduced in FSharp.Core. 
On the other hand, I think it's a bit quirky to have generic `choose` which is applied as `chooseValues` for key-value types. In most (all?) cases, F#+ generic functions are applied to functions with the same name.
Another possibility is to create generic `chooseValues` and apply it only to key-value types. I feel it's less elegant but would certainly work.

wallymathieu:
I agree that `choose` should probably be named as `chooseBy`. I'm a bit unsure about what makes sense for choose for key value collections such as dictionaries and maps. The implementation seen above makes sense from one point of view.

gusty:
> On the other hand, I think it's a bit quirky to have generic `choose` which is applied as `chooseValues` for key-value types. In most (all?) cases, F#+ generic functions are applied to functions with the same name.

There are more cases like this. FSharp.Core is not a type classes based library, so there are plenty of inconsistences between modules. Here we focus more on functionality than in names, when we have a generic function it attempts to match the name of a common non-generic function, but if that common function is inconsistently defined in one module we use that definition.

An example is `map`, which in FSharp.Core for `Map` is acting on keys as well, which in fact correspond to our `mapi`, and we added `mapValues` which correspond to the generic `map`.

TLDR: We're not trying to fix FSharp.Core, so we do our best without becoming invasive.

src/FSharpPlus/Extensions/Dict.fs
```diff
@@ -73,6 +73,19 @@ module Dict =
             | None    -> ()
         dct :> IDictionary<'Key, 'U>
 
+    /// <summary>Applies given function to each value of the given dictionary.</summary>
+    /// <param name="f">The mapping function.</param>
+    /// <param name="x">The input dictionary.</param>
+    ///
+    /// <returns>Returns dictionary with values x for each dictionary value where the function returns Some(x).</returns>
+    let choose f (x: IDictionary<'Key, 'T>) =
```
wallymathieu, line 81: I think this makes sense from the point of view that dictionaries are simply collections indexed by a key.

tests/FSharpPlus.Tests/General.fs
```diff
@@ -690,6 +690,20 @@ module Collections =
         let rl = ReadOnlyListOnlyIndex [1..10]
         Assert.AreEqual (2, nth 1 rl)
 
+    [<Test>]
+    let choose () =
```
wallymathieu, line 694: ❤️

wallymathieu (approved):

wallymathieu:
My feeling is that if you look at Map and Dictionary as simply collection types indexed by something else than ordinal, I think this change makes sense. Though that also raises the question if choose should be defined more generally?

wallymathieu:
I think the PR from a code point of view looks fine. The tricky question is how it could interact with decisions in future F# Core. Would code subtly break due to ordering of imports?

src/FSharpPlus/Extensions/Dict.fs
gusty: On one side I like to call it `choose` but I fear the risk of having to deal with an eventual inclusion of a `choosei` function in F# core, called `choose`.

This will need more feedback from users.

So, I would say, if you want to merge it now, please rename them to `chooseValues` so we stay in the safe side. As I said before, we can always move to `choose` if we firmly decide to.

gusty (changes requested):

gusty (approved):

PR #370, opened 2020-10-10
Add zipShortest function and make generic Zip safe

3Rafal:
Added safe zip function `zipShortest`. 
It's behavior is equivalent to Haskell `zip` which I think is quite reasonable.
Naming is already used in .NET ecosystem in MoreLinq C# library [[Source](https://morelinq.github.io/2.10/ref/api/html/M_MoreLinq_MoreEnumerable_ZipShortest__3.htm)]

### DIFF
--- a/src/FSharpPlus/Data/NonEmptySeq.fs
+++ b/src/FSharpPlus/Data/NonEmptySeq.fs
@@ -463,14 +463,6 @@ module NonEmptySeq =
     ///
     /// <returns>The result sequence.</returns>
     let zip3 (source1: NonEmptySeq<_>) (source2: NonEmptySeq<_>) (source3: NonEmptySeq<_>) = Seq.zip3 source1 source2 source3 |> unsafeOfSeq
-
-    /// <summary>
-    /// Zip safely two sequences. If one sequence is shorter, excess elements are discarded from the right end of the longer sequence. 
-    /// </summary>
-    /// <param name="a1">First input sequence.</param>
-    /// <param name="a2">Second input sequence.</param>
-    /// <returns>Sequence with corresponding pairs of input sequences.</returns>
-    let zipShortest (source1: NonEmptySeq<_>) (source2: NonEmptySeq<_>) = Seq.zipShortest source1 source2 |> unsafeOfSeq
     
     /// <summary>Applies the given function to each element of the NonEmptySequence and concatenates all the
     /// results.</summary>
--- a/src/FSharpPlus/Data/NonEmptySeq.fs
+++ b/src/FSharpPlus/Data/NonEmptySeq.fs
@@ -463,7 +463,7 @@ module NonEmptySeq =
     ///
     /// <returns>The result sequence.</returns>
     let zip3 (source1: NonEmptySeq<_>) (source2: NonEmptySeq<_>) (source3: NonEmptySeq<_>) = Seq.zip3 source1 source2 source3 |> unsafeOfSeq
-    
+
     /// <summary>Applies the given function to each element of the NonEmptySequence and concatenates all the
     /// results.</summary>
     ///
--- a/src/FSharpPlus/Data/NonEmptySeq.fs
+++ b/src/FSharpPlus/Data/NonEmptySeq.fs
@@ -464,6 +464,14 @@ module NonEmptySeq =
     /// <returns>The result sequence.</returns>
     let zip3 (source1: NonEmptySeq<_>) (source2: NonEmptySeq<_>) (source3: NonEmptySeq<_>) = Seq.zip3 source1 source2 source3 |> unsafeOfSeq
 
+    /// <summary>
+    /// Zip safely two sequences. If one sequence is shorter, excess elements are discarded from the right end of the longer sequence. 
+    /// </summary>
+    /// <param name="a1">First input sequence.</param>
+    /// <param name="a2">Second input sequence.</param>
+    /// <returns>Sequence with corresponding pairs of input sequences.</returns>
+    let zipShortest (source1: NonEmptySeq<_>) (source2: NonEmptySeq<_>) = Seq.zipShortest source1 source2 |> unsafeOfSeq
+    
     /// <summary>Applies the given function to each element of the NonEmptySequence and concatenates all the
     /// results.</summary>
     ///
--- a/src/FSharpPlus/Extensions/ResizeArray.fs
+++ b/src/FSharpPlus/Extensions/ResizeArray.fs
@@ -86,3 +86,15 @@ module ResizeArray =
         let (x, y) = ResizeArray (), ResizeArray ()
         Array.iter (mapper >> function Choice1Of2 e -> x.Add e | Choice2Of2 e -> y.Add e) source
         x.ToArray (), y.ToArray ()
+    
+    
+    /// <summary>
+    // Zip safely two ResizeArrays. If one ResizeArray is shorter, excess elements are discarded from the right end of the longer ResizeArray. 
+    /// </summary>
+    /// <param name="a1">First input ResizeArray.</param>
+    /// <param name="a2">Second input ResizeArray.</param>
+    /// <returns>ResizeArray with corresponding pairs of input ResizeArrays.</returns>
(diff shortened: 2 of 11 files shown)


adz:
I like your idea!

I thought this was how List.zip in core worked already, but I was wrong. It throws exception if different sized lists are zipped.

Unfortunately, F#+ seems a little inconsistent:
 * extensions for Enumerator, Dictionary, IReadOnlyDictionary, Map implement `zip` just like your zipShortest (i.e. it stops when one enumerator stops, and skips over missing keys for map, etc). 
 * NonEmptyList.fs and NonEmptySeq.fs in FSharpPlus/Data directly call List.zip / Seq.zip, so these throw.

So for completeness, we should probably implement `zipShortest` for NonEmptyList, NonEmptySeq.

I'm not sure if it's worth also implementing `zipShortest` as `zip` in the other cases?

adz:
Oh now I notice the other pr!

Well seems to me still makes sense to implement for non empty list... tryZip could be a separate thing.

3Rafal:
@adz , you're right. I've added implementation for `nelist` and `neseq` as you suggested.
When it comes to `tryZip`, I'm not convinced it is a good idea, so I won't implement it now.

gusty:
> Unfortunately, F#+ seems a little inconsistent:

Again it worth clarifying is not our goal to fix FSharp.Core 
If they took the unfortunate decision to make lists and arrays to fail when zipping we have to honor it, if we have a nelist it can't behave differently to the normal list, but same goes for neseq.

So these inconsistencies comes from F# core and we can't fix them. We should instead create different function which behave in a consistent way, take for instance the example of `take` and `skip`, they fail but we introduced other functions `drop` and `limit` that don't.

gusty:
We could however, define our generic `zip` as `zipShortest`, that would solve inconsistencies at the generic level.
Same would apply for `map2`.

But I definitely like the `zipShortest` in the extensions.

3Rafal:
@gusty , I've introduced changes according to your remarks. now generic `zip` is using safe functions only.

src/FSharpPlus/Extensions/Seq.fs
```diff
@@ -78,6 +78,18 @@ module Seq =
                     members <- ResizeArray ()
                     members.Add e.Current
             yield g, members }
+    
+    /// <summary>
```
gusty, line 82: Should we include `zipShortest` even in types like `seq` and `neseq` which already have `zip` with that behavior?

I think this is not necessary.

gusty (changes requested):

src/FSharpPlus/Operators.fs
```diff
@@ -93,10 +93,10 @@ module Operators =
     // Un-zips (un-tuple) two functors.
     let inline unzip (source: '``Functor<'T1 * 'T2>``) = Unzip.Invoke source : '``Functor<'T1>`` * '``Functor<'T2>``
 
-    // Zips (tuple) two functors.
+    // Zips safely two collections. If one collection is shorter, excess elements are discarded from the right end of the longer collection.
```
gusty, line 96: With this comment you are downgrading a function which operates with 2 functors to a function which operates only with 2 collection functors :)  

I see your intention of clarifying the "shortest" behavior and I share it.

Maybe, leave the original comment and add yours in remarks, something like "for collections, if one collection is shorter ..."

gusty (changes requested):

src/FSharpPlus/Extensions/Seq.fs
3Rafal: True, removed

src/FSharpPlus/Operators.fs
3Rafal: Changed as suggested. :-)

src/FSharpPlus/Data/NonEmptySeq.fs
```diff
@@ -463,7 +463,7 @@ module NonEmptySeq =
     ///
     /// <returns>The result sequence.</returns>
     let zip3 (source1: NonEmptySeq<_>) (source2: NonEmptySeq<_>) (source3: NonEmptySeq<_>) = Seq.zip3 source1 source2 source3 |> unsafeOfSeq
-
+
```
gusty, line 466: Just one more thing. Is there a way you can remove this change? I mean, in such a way that this file is not touched at all by this PR. Sorry for nitpicking, it's just to keep our history (which is about 10 years old) clean.

gusty (changes requested):

src/FSharpPlus/Data/NonEmptySeq.fs
3Rafal: Sorry, my IDE sometimes messes up whitespace. Fixed :-)

src/FSharpPlus/Extensions/ResizeArray.fs
```diff
@@ -86,3 +86,15 @@ module ResizeArray =
...
+    /// <summary>
+    /// Zip safely two ResizeArrays. If one ResizeArray is shorter, excess elements are discarded from the right end of the longer ResizeArray. 
+    /// </summary>
+    /// <param name="a1">First input ResizeArray.</param>
+    /// <param name="a2">Second input ResizeArray.</param>
+    /// <returns>ResizeArray with corresponding pairs of input ResizeArrays.</returns>
+    let zipShortest (a1: ResizeArray<'T1>) (a2: ResizeArray<'T2>) =
+        let len = min a1.Count a2.Count
+        [| for i in 0..(len-1) -> a1.[i], a2.[i] |]
+        |> ResizeArray
```
gusty, line 100: I will accept it as it is, but not sure if this is the most efficient implementation in terms of allocations.

gusty (approved):

src/FSharpPlus/Operators.fs
```diff
@@ -94,9 +94,10 @@ module Operators =
     let inline unzip (source: '``Functor<'T1 * 'T2>``) = Unzip.Invoke source : '``Functor<'T1>`` * '``Functor<'T2>``
 
     // Zips (tuple) two functors.
+    // For collections, if one collection is shorter, excess elements are discarded from the right end of the longer collection. 
     let inline zip (source1: '``ZipFunctor<'T1>``) (source2: '``ZipFunctor<'T2>``) : '``ZipFunctor<'T1 * 'T2>`` = Zip.Invoke source1 source2
```
adz, line 98: @rafal I can see this is where your doc probs initiated.

These comments should be `///` not `//. If you don't mind, could you fix these also?

src/FSharpPlus/Extensions/ResizeArray.fs
3Rafal: Changed it to efficient version both here and in ResizeArray.

src/FSharpPlus/Operators.fs
3Rafal: No problem :-)

src/FSharpPlus/Extensions/ResizeArray.fs
```diff
@@ -86,3 +86,15 @@ module ResizeArray =
...
         x.ToArray (), y.ToArray ()
+    
+    
+    /// <summary>
+    // Zip safely two ResizeArrays. If one ResizeArray is shorter, excess elements are discarded from the right end of the longer ResizeArray. 
+    /// </summary>
+    /// <param name="a1">First input ResizeArray.</param>
+    /// <param name="a2">Second input ResizeArray.</param>
+    /// <returns>ResizeArray with corresponding pairs of input ResizeArrays.</returns>
+    let zipShortest (a1: ResizeArray<'T1>) (a2: ResizeArray<'T2>) =
```
3Rafal, line 97: @gusty , I wonder if the naming is right. FSharp.Core doesn't have a `zip` function for ResizeArray, so we could use just `zip` here. However, `FSharpx.Collections` have `ResizeArray.zip` that fails when arrays are of different length.

gusty: You are right, let's stay on more time on the safe side: leave as it is now.

PR #371, opened 2020-10-10
Add (some) categories to Operators.fs

adz:
This is incomplete, but adds categories to operators for better layout in docs. Note you have to use xml docstrings for this to work.


adz:
Merged in separate PR

PR #372, opened 2020-10-10
Update to latest F# Formatting

wallymathieu:
- [x] upgrade to latest F# formatting
- [x] make sure that reference docs builds
- [x] normalize template location with what fsdocs program tool expects

The goal is to be able to upgrade to the latest F# Formatting as seen in https://github.com/fsprojects/FSharpPlus/pull/369 but keep the possibility to deploy docs and use the plantuml diagram.


docsrc/content/_template.html
```diff
@@ -4,16 +4,16 @@
     <meta charset="utf-8"/>
```
wallymathieu, line 4: Note that this template does not work well, since It's not the one that @adz has fixed. This is simply the minimal effort conversion of template.cshtml to html.

wallymathieu:
I've rendered the docs here:
https://wallymathieu.github.io/FSharpPlus/
Note that the above rendering is using the template by @adz https://github.com/wallymathieu/FSharpPlus/tree/wally-upgrade-formatting-adz

docsrc/content/abstraction-monad.fsx
```diff
@@ -193,7 +194,7 @@ let getValidPassword : ResultT<_> =
 let askPassword = monad {
     do! lift <| putStrLn "Insert your new password:"
     let! value = getValidPassword
-    do! lift <| putStrLn "Storing in database..."
+    //do! lift <| putStrLn "Storing in database..."
```
adz, line 197: I'm guessing this caused an error? ..or maybe it was unintentional?

adz: Checking locally, I got:

```
   docsrc/content/abstraction-monad.fsx: error(197,8)-(197,53) Type constraint mismatch when applying the default type 'obj' for a type inference variable. The type 'obj' does not support the operator 'Lift' Consider adding further type constraints
```

Adding type constraint didn't help -- I got "Value restriction".
Changing `lift` -> `liftAsync` doesn't help either.

I resolved it by passing unit:

```
let askPassword () = monad {
...
```

adz (approved):
I think we should merge. Some of the doc issues may have been present before, or not, but we can resolve separately. They still generate a doc.

docsrc/content/abstraction-monad.fsx
wallymathieu: We should apply your fix to the docs.

PR #373, opened 2020-10-10
Introduce map2Shortest function

3Rafal:
It's a similar idea to already introduced `zipShortest`.


gusty (approved):
At some point we should add a generic `map2`, which will use these functions. Currently we have `lift2` which in many cases is the same as `map2`, but not in collections, except the `ZipList`.

Issue #44, opened 2017-03-26
Bind for Expr<'T> (Quotation Monad)


gusty:
After a short but interesting discussion with @eiriktsarpalis I thought it might be interesting to add a ``Bind`` implementation for ``Expr<'t>`` which allows to combine functions returning quotations.

One open question is how the quotation will be evaluated, however I did a [draft some years ago](https://gist.github.com/gusty/9f367de5ac8393979969) which uses the splice operator to signify that there is an evaluation at that point, together with a ``runUntyped`` function which accepts any quotation evaluator and makes sure all splices are evaluated properly.

It might be worth adding a specific implementation of quotation evaluator to this library, but it has to be very efficient. At the moment I wrote that code there were no efficient implementations around.

PR #374, opened 2020-10-11
Quotation Monad

gusty:
This will implement #44 

A quotation monad allow us to compose function returning expressions.

At the end of the composition we get a single expression, internally containing markers where an evaluate function needs to run  a sub-expression.

Since F# doesn't support the "run" expression, we can't directly run this expressions in existing quotation evaluators, but we include an `Expr.run` function which takes any normal quotation evaluator and takes care of handling the internal marker. So, instead of doing `evaluator expr` we do `Expr.run evaluator expr`.

### DIFF
--- a/src/FSharpPlus/Control/Monad.fs
+++ b/src/FSharpPlus/Control/Monad.fs
@@ -35,6 +35,8 @@ type Bind =
     static member        (>>=) (source             , k: 'T -> _           ) = Result.bind k source                                          : Result<'U,'E>
     static member        (>>=) (source             , k: 'T -> _           ) = Choice.bind k source                                          : Choice<'U,'E>
 
+    static member        (>>=) (source: Expr<'T>   , f: 'T -> Expr<'U>    ) = Expr.bind f source                                            : Expr<'U>
+
     static member (>>=) (source: Map<'Key,'T>, f: 'T -> Map<'Key,'U>) = Map (seq {
                    for KeyValue(k, v) in source do
                        match Map.tryFind k (f v) with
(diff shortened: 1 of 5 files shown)


cannorin (commented):
I experimented around it a lot and I think I understand why the current implementation looks like this. I think this is one of the best we can have for a "quotation monad", but here is my two cents.

Here, we can't make use of the created `Expr<'a>` value for anything but executing it, because it contains `SpliceExpression`s which make the expression illegal and have to be removed (by eval).  This makes me wondering how this "monad" can be useful, since I suppose most people would want to use the computation expression to create a valid `Expr<'a>`  value that can be consumed by other libraries (without actually running it), but it isn't possible (this is not a problem in MetaOCaml, since it's all about running the code and there is no API to directly manipulate the AST).

Since a F# code can have side effects anywhere, we can't easily define `m >>= f` without breaking the semantics. We have to syntactically inspect the `f` and replace every occurrence of the argument with `x` (this is not a problem in Template Haskell since it uses IO monad so that can be done safely). Luckily we can do this for the computation expression binding `let! x = m in body` by using `[<ReflectedDefinition>]` since it is always a lambda abstraction, but this means generic `>>=` operator and generic `monad` builder are no more usable for the quotation monad.

So I think we have several options:
* Wrap the `Expr<'a>` type (in `WrappedExpr<'a>`, for example) and make it a monad.
  - Then we define `val WrappedExpr.run : (Expr -> obj) -> WrappedExpr<'a> -> 'a`.
  - And `val WrappedExpr.toExpr : WrappedExpr<'a> -> Expr<'a>`. This requires the syntactical replacement described above.
* Create a specialized builder like `quotation { ... }` for the purpose described above.
  - We also have to implement the syntactical replacement for this.
  - Then we may or may not make `Expr<'a>` a monad.
* Do both of them.
* Do neither and keep it as is.

Also, I found an interesting paper around this topic: https://www.semanticscholar.org/paper/On-Meta-Programming-and-Code-Generation-in-F-Larjani/ca1a48245a767992f4461fe2e66217e7a4fa83d6

Issue #327, opened 2020-05-22
Add tryLast for Foldable


cmeeren:
I see that `tryHead` is implemented, but `tryLast` is not. It would nicely complement the existing `tryHead`.

wallymathieu:
My personal opinion is that it would make sense to add tryLast since there is `List.tryLast`.

gusty:
How would it work with infinite sequences? I don't think we can make it safe in that case.

wallymathieu:
That's true. That's one thing that speaks against adding it.

cmeeren:
You have `Seq.tryLast` in F# code, so it doesn't seem to have stopped them from adding it.

wallymathieu:
`Seq.tryLast` in f# core is probably to be compatible with common LINQ usage in .net.

cmeeren:
I don't see how that changes anything; please enlighten me if I am missing something.

wallymathieu:
I imagine that many usages of `seq` are more intended to wrap a finite collection rather than a potentially unbounded collection. In c# you sometimes have `seq` hiding a generator:

```c#
public IEnumerable<T> GetUnboundedResults()
{
        while(true)
        {
            var result = GenerateNextResult();
            yield return result;
        }
}
```

I've seen this pattern once or twice, but mostly people use `IEnumerable` in order to avoid leaking a specific collection implementation that they don't intend for you to modify (without creating a new collection based on it).

gusty:
So, `tryLast` will be aimed to "fail" on empty structures, right?

cmeeren:
Shouldn't `tryLast` return `None` for empty structures? That's what `Seq.tryLast`, `List.tryLast` etc. does.

gusty:
Yes, that's what I meant with "fail", I was referring to the pure-failure, which in this case is `None`.

gusty:
Implemented in #375

PR #375, opened 2020-10-11
Add tryLast to Foldable

3Rafal:
As discussed in RFC #327

### DIFF
--- a/tests/FSharpPlus.Tests/General.fs
+++ b/tests/FSharpPlus.Tests/General.fs
@@ -7,6 +7,7 @@ open System.Collections.ObjectModel
 open FSharpPlus
 open FSharpPlus.Data
 open FSharpPlus.Control
+open FSharpPlus.Internals
 open NUnit.Framework
 open Helpers
 open FSharpPlus.Math.Applicative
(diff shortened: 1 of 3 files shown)


tests/FSharpPlus.Tests/General.fs
```diff
@@ -872,6 +873,39 @@ module Foldable =
...
+        let a                = tryLast [|1|]
+        let a': int option   = tryLast [||]
+        areEqual a (Some 1)
+        areEqual a' None
+
+        let nes              = tryLast <| NonEmptySeq.ofList [1;2]
+        areEqual nes (Some 2)
+
+        let i                = tryLast <| Id.create 4
+        areEqual i (Some 4)
```
gusty, line 897: `Id` is a dummy type we use internally to create ambiguity in the overload resolution, so better remove it from the tests, it could create confusion and there is not point at all in testing it.

gusty (changes requested):
Looks goods, thanks. Just one detail.

tests/FSharpPlus.Tests/General.fs
3Rafal: Ok, removed

gusty (approved):

PR #377, opened 2020-10-13
DRY in Functor.Map

3Rafal:
When working on Functor.Map2, I've found some duplications/redundancies in Functor.Map.
Changed it and added missing tests for Async.


gusty:
Thanks.
Please note that `Map2` is mainly `Lift2`, but some collections accept the alternative implementation of the zipList behavior.
For those, we can define `Map2` but it would be for only for collections I think.

3Rafal:
> Please note that Map2 is mainly Lift2, but some collections accept the alternative implementation of the zipList behavior.
For those, we can define Map2 but it would be for only for collections I think.

Ok, this work will be done on different PR, because I think this is self-contained. So I guess we can merge this one now.

gusty (approved):

PR #378, opened 2020-10-20
+Example from F# Online

### LINKED ISSUE #326: Include missing examples from f# meetup
### Description

A use of reader monad that could be included in the documentation was shown in: 
https://www.youtube.com/watch?v=pxJCHJgG8ws
by @starychfojtu

```f#
open System
open FSharpPlus
open FSharpPlus.Data

type IUserRepository =
    abstract GetUser : email : string -> string

type IShoppingListRepository =
    abstract Add : shoppingList : string list -> string list

let getUser email =
    Reader(fun (env : #IUserRepository) -> env.GetUser email)

let addShoppingList shoppingList =
    Reader(fun (env : #IShoppingListRepository) -> env.Add shoppingList)

type Env() =
    interface IUserRepository with
        member this.GetUser email =
                "Sandeep"    
    interface IShoppingListRepository with
            member this.Add shoppingList =
                shoppingList
```

wallymathieu:
This will fix #326

### DIFF
--- a/docsrc/content/type-reader.fsx
+++ b/docsrc/content/type-reader.fsx
@@ -89,3 +89,40 @@ and
                                         let! name = resolve t
                                         let! value = resolve d
                                         return (name,value) }
+
+
+(**
+Another usage of the Reader monad is an alternative to dependency injection or currying
+in order to pass around dependencies. The below code comes from [F# Online - Josef Starýchfojtů - FSharpPlus - Advanced FP concepts in F#](https://www.youtube.com/watch?v=pxJCHJgG8ws). You can find the presenter on github as [@starychfojtu](https://github.com/starychfojtu).
+
+Why would you want to do this style?
+
+- When you want to pass around a single environment instead of using dependency injection.
+
+Why wouldn't you want to use this style?
+
+- The downside of this style is that it supposes that your environment is relatively immutable. If you have different lifetimes for different implementation classes dependency injection frameworks can be easier to use.
+*)
+open System
+open FSharpPlus
+open FSharpPlus.Data
+
+type IUserRepository =
+    abstract GetUser : email : string -> string
+
+type IShoppingListRepository =
+    abstract Add : shoppingList : string list -> string list
+
+let getUser email =
+    Reader(fun (env : #IUserRepository) -> env.GetUser email)
+
+let addShoppingList shoppingList =
+    Reader(fun (env : #IShoppingListRepository) -> env.Add shoppingList)
+
+type Env() =
+    interface IUserRepository with
+        member this.GetUser email =
+                "Sandeep"
+    interface IShoppingListRepository with
+            member this.Add shoppingList =
+                shoppingList
--- a/docsrc/content/type-reader.fsx
+++ b/docsrc/content/type-reader.fsx
@@ -20,12 +20,63 @@ Examples
 
 
 #r @"../../src/FSharpPlus/bin/Release/net45/FSharpPlus.dll"
+
+(**
+One usage of the Reader monad is an alternative to dependency injection or currying
+in order to pass around dependencies. The below code comes from [F# Online - Josef Starýchfojtů - FSharpPlus - Advanced FP concepts in F#](https://www.youtube.com/watch?v=pxJCHJgG8ws). You can find the presenter on github as [@starychfojtu](https://github.com/starychfojtu).
+
+Why would you want to do this style?
+
+- When you want to pass around a single environment instead of using dependency injection.
+
+Why wouldn't you want to use this style?
+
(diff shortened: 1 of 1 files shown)


docsrc/content/type-reader.fsx
```diff
@@ -89,3 +89,40 @@ and
...
+open FSharpPlus.Data
+
+type IUserRepository =
+    abstract GetUser : email : string -> string
+
+type IShoppingListRepository =
+    abstract Add : shoppingList : string list -> string list
+
+let getUser email =
+    Reader(fun (env : #IUserRepository) -> env.GetUser email)
```
adz, line 117: What does the '#' mean there?

wallymathieu: It means that you don't need exactly a `IUserRepository` but a type that implements `IUserRepository`. The name for this is [flexible type](https://docs.microsoft.com/en-us/dotnet/fsharp/language-reference/flexible-types). Perhaps we should write some text explaining what the example does?

docsrc/content/type-reader.fsx
```diff
@@ -89,3 +89,40 @@ and
...
+
+let addShoppingList shoppingList =
+    Reader(fun (env : #IShoppingListRepository) -> env.Add shoppingList)
+
+type Env() =
```
adz, line 122: Maybe call this "MockDataEnv", or add a comment to say "This is how an environment could be constructed"

docsrc/content/type-reader.fsx
```diff
@@ -89,3 +89,40 @@ and
                                         let! name = resolve t
                                         let! value = resolve d
                                         return (name,value) }
+
+
+(**
+Another usage of the Reader monad is an alternative to dependency injection or currying
+in order to pass around dependencies. The below code comes from [F# Online - Josef Starýchfojtů - FSharpPlus - Advanced FP concepts in F#](https://www.youtube.com/watch?v=pxJCHJgG8ws). You can find the presenter on github as [@starychfojtu](https://github.com/starychfojtu).
```
adz, line 96: I would put this before the example above, and say "One usage of..."

docsrc/content/type-reader.fsx
```diff
@@ -89,3 +89,40 @@ and
...
+
+Why would you want to do this style?
+
+- When you want to pass around a single environment instead of using dependency injection.
+
+Why wouldn't you want to use this style?
+
+- The downside of this style is that it supposes that your environment is relatively immutable. If you have different lifetimes for different implementation classes dependency injection frameworks can be easier to use.
```
adz, line 104: Also note above the consideration of the 'State' related type.

docsrc/content/type-reader.fsx
```diff
@@ -89,3 +89,40 @@ and
...
+let addShoppingList shoppingList =
+    Reader(fun (env : #IShoppingListRepository) -> env.Add shoppingList)
+
+type Env() =
+    interface IUserRepository with
+        member this.GetUser email =
+                "Sandeep"
+    interface IShoppingListRepository with
+            member this.Add shoppingList =
+                shoppingList
```
adz, line 128: It would be nice to show how it all fits together: where the actual repository is instantiated, and passed through.

Also, I'm not sure about adding "shoppingList" as an example.
I guess it's supposed to be like a cart? or you can have multiple lists or something.

... this is just his example, but using "env.Add" seems a bit confusing. What's it adding?

I think the main point is to show two different 'repositories' and functions using them, then how the reader can weave them together. That is good, but maybe we could tweak it a bit?

BTW -- Good idea to add the example. :+1:

docsrc/content/type-reader.fsx
wallymathieu: How do you mean? That you could instead weave lifetime scoped entities in 'State'?

docsrc/content/type-reader.fsx
wallymathieu: I've tried to adjust the example a bit.

docsrc/content/type-reader.fsx
adz: You said: `it supposes that your environment is relatively immutable.` State type allows you to set and get, effectively mutating.

...and above says:

 - [State](type-state.html): Similar, but it allows you to modify the environment.

docsrc/content/type-reader.fsx
adz: Looks good!

adz (approved):
Looks good to me.

gusty (approved):

docsrc/content/type-reader.fsx
wallymathieu: Yes, that's true. Though the lifetime issues probably applies to State as well. I guess the solution when you have a reader environment would be to have a proxy so that you control the lifetime separate from the normal business logic flow?

PR #379, opened 2020-10-21
Simplify Lens docs

3Rafal:
I think Lens example is clearer when using update expression. 
Rewriting all record values by hand seems like suboptimal practice, that doesn't scale for more record fields.
Also, I've added examples of updating a value through lens.

### DIFF
--- a/docsrc/content/lens.fsx
+++ b/docsrc/content/lens.fsx
@@ -45,7 +45,7 @@ type Book =
 
 module Book =
     let inline _author f b =
-        f b.Author <&> fun a' -> { b with Author = a' }
+        f b.Author <&> fun a -> { b with Author = a }
 
     let inline _authorName b = _author << Person._name <| b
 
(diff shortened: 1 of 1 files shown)


docsrc/content/lens.fsx
```diff
@@ -30,28 +30,38 @@ open System
...
+    
 // read book author name:
 let authorName1 = view Book._authorName rayuela
 //  you can also write the read operation as:
 let authorName2 = rayuela ^. Book._authorName
 
+// write value through a lens
+let book1 = setl Book._authorName "William Shakespear" rayuela
+// update value
+let book2 = over Book._authorName (fun x -> x.ToUpper()) rayuela
```
gusty, line 64: here we can use our `String.toUpper` extension.

gusty (approved):

docsrc/content/lens.fsx
3Rafal: true, changed

PR #380, opened 2020-10-21
Use list instead of array for the generic implementation of sequence

Arshia001:
First off, let me say that this library is AMAZING. My mind is completely blown. Nice. Seriously.

Now, on to the matter at hand:

The generic implementation of sequence uses arrays, specifically `Array.append`, which is linear in the length of the array. This means the entire operation is of exponential order. Sequencing a 1_000_000 element array on my i9 9900K takes longer than I have the patience to wait (more than a minute).

With that said, this pull request changes the implementation to use the usual list cons operation, which happens in constant time, with a linear List.rev at the end, which brings the order of the entire operation down to linear. A 10_000_000 item array (10 times as many elements) is now sequenced in under 4 seconds (about 3.7 seconds, to be more precise). I didn't test the memory consumption and GC, but I believe those will have improved as well, since we aren't allocating a million arrays any more.

### DIFF
--- a/src/FSharpPlus/Control/Traversable.fs
+++ b/src/FSharpPlus/Control/Traversable.fs
@@ -18,14 +18,14 @@ type Sequence =
 
     [<EditorBrowsable(EditorBrowsableState.Never)>]
     static member inline ForInfiniteSequences (t: seq<_>, isFailure, conversion) =
-        let add x y = Array.append x [|y|]
+        let add x y = y :: x
         let mutable go = true
-        let mutable r = result [||]
+        let mutable r = result []
         use e = t.GetEnumerator ()
         while e.MoveNext () && go do
             if isFailure e.Current then go <- false
             r <- Map.Invoke add r <*> e.Current
-        Map.Invoke conversion r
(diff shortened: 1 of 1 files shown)


src/FSharpPlus/Control/Traversable.fs
```diff
@@ -18,14 +18,14 @@ type Sequence =
...
     static member inline ForInfiniteSequences (t: seq<_>, isFailure, conversion) =
-        let add x y = Array.append x [|y|]
+        let add x y = y :: x
         let mutable go = true
-        let mutable r = result [||]
+        let mutable r = result []
         use e = t.GetEnumerator ()
         while e.MoveNext () && go do
             if isFailure e.Current then go <- false
             r <- Map.Invoke add r <*> e.Current
```
wallymathieu, line 27: Yes, makes sense.

gusty:
I think this was the original implementation, then I switched to array in order to get better performance but since the array have to be constructed step by step it gets even worst.

Thanks for taking care of this !

PR #382, opened 2020-10-22
Add operator doc categories

adz:
This adds categories with indexing to operators. This produces a nice table of contents which matches the banner comments in the source.

There is an oddity with the new FSharp Formatting, for the "</" operator... it seems to get interpreted as HTML, and doesn't display. I've linked to the more detailed description:

image


gusty (approved):

wallymathieu:
The build failure seems to be due to the doc tool assuming that the files folder contains static files. Should probably be easy to fix.

wallymathieu:
I can do the fix here @adz in order for us to be able to merge this PR.

adz:
If you have the time - go ahead, thanks @wallymathieu I won't get back to it for a few days

wallymathieu (approved):
👍

PR #383, opened 2020-10-22
Simplify and format Prism docs

3Rafal:
Simplify Prism docs using update expression.
Capitalize record field names according to MS F# style guide.
Format code.
Add remark on polymorphic update and `with` expression incompatibility.


gusty (approved):

wallymathieu:
Thanks for the contribution @3Rafal! It's nice that we have some attention to the docs 😃

PR #389, opened 2020-11-01
Sample usage of C# default interface implementation together with tryParse operator

wallymathieu:


tests/FSharpPlus.Tests/General.fs
```diff
@@ -2237,7 +2244,7 @@ module Sequences =
           let! z = seq [1..50]
           let! x = seq [1..z]
           let! y = seq [x..z]
-          if (x*x + y*y = z*z) then return (x, y, z)}
+          if (x*x + y*y = z*z) then return (x, y, z) else () }
```
wallymathieu, line 2247: This looks like a change in F# 5

tests/FSharpPlus.Tests/General.fs
```diff
@@ -1784,6 +1784,7 @@ module MonadTransformers =
         let _ = put initialState : ChoiceT<State<int, Choice<unit,string>>>
 
         ()
+#if !NETSTANDARD3_0
```
wallymathieu, line 1787: These examples does not compile with F# 5, so are probably regressions

wallymathieu:
Note the changes when compiling with F# 5. The type inference for CE builders seems to be a breaking change, unsure if it's intentional. @dsyme 
Note that these are the same changes became evident for in the PR that updated F# formatting:
https://github.com/fsprojects/FSharpPlus/pull/372

gusty:
I created a standalone repro and submitted it https://github.com/dotnet/fsharp/issues/10379

tests/CSharpLib/CustomerId.cs
```diff
@@ -23,4 +26,17 @@ public static bool TryParse(string value,out CustomerId id)
             return false;
         }
     }
+#if NETSTANDARD2_1 
+    public interface ICustomerId{
```
gusty, line 30: We don't have guidelines for C# in this projects, but at least could you leave a space after `;` and surround with spaces `{`, , `}` and `=`, which I think is kind of standard in both languages?

gusty (approved):
Cool, you can merge it, after the suggested change.

tests/CSharpLib/CustomerId.cs
wallymathieu: As usual I'm a bit sloppy about indentation. Thanks!

PR #394, opened 2020-11-13
Make NonEmptySeq.unsafeOfSeq public

cannorin:
`NonEmptySeq.ofSeq` uses `Seq.isEmpty` to check if the sequence is actually non-empty.

But `Seq.isEmpty` evaluates the first element of the sequence to do that, so using `ofSeq` may trigger the side effects.

```fsharp
let s = seq {printfn "a"; 1; printfn "b"; 2; printfn "c"; 3; }
let ns = NonEmptySeq.ofSeq s // prints "a"
```

This can be problematic when a user know the sequence is non-empty and does not want to trigger the said side effects. This PR makes the internal function `unsafeOfSeq` public for that purpose, with some document comments to indicate that user should normally use `ofSeq`.

TODO:
- [ ] Add tests?

### DIFF
--- a/src/FSharpPlus/Data/NonEmptySeq.fs
+++ b/src/FSharpPlus/Data/NonEmptySeq.fs
@@ -15,11 +15,24 @@ type NonEmptySeq<'T> =
 type neseq<'t> = NonEmptySeq<'t>
 
 module NonEmptySeq =
-    let internal unsafeOfSeq (x: _ seq) =
+    /// <summary>Builds a non empty sequence from the given sequence.</summary>
+    /// <param name="seq">The input sequence.</param>
+    /// <returns>Non empty sequence containing the elements of the list.</returns>
+    /// <remarks>
+    ///   **This function does not check whether the sequence is actually non empty or not.**
+    /// 
+    ///   Use this function only if you are sure that the sequence is not empty and
+    ///   you don't want to evaluate the first element of the sequence which would cause a
+    ///   side effect.
+    ///  
+    ///   Otherwise, always use `ofSeq`. 
+    /// </remarks>
+    /// <seealso cref="ofSeq" />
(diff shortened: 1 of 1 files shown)


gusty (approved):
Looks good. For testing we can use the example already mentioned but changing `printfn` for `sideEffects.add` so we can check them at the end.

src/FSharpPlus/Data/NonEmptySeq.fs
```diff
@@ -15,11 +15,24 @@ type NonEmptySeq<'T> =
...
+    ///   **This function does not check whether the sequence is actually non empty or not.**
+    /// 
+    ///   Use this function only if you are sure that the sequence is not empty and
+    ///   you don't want to evaluate the first element of the sequence which would cause a
+    ///   side effect.
+    ///  
+    ///   Otherwise, always use `ofSeq`. 
+    /// </remarks>
+    /// <seealso cref="ofSeq" />
+    let unsafeOfSeq (seq: _ seq) =
```
3Rafal, line 31: Don't mean to be nitpicky, but I wonder if it's a good idea to use one of the most popular FSharp.Core function name as a parameter name.
FSharp.Core uses convention of `source` as a parameter name for seq to avoid shadowing.

gusty: Is not that bad, I mean not that uncommon in F# code I think.

But I agree in that it’s better to use another name, I think FSharp.Core (implicit) convention is to call it `source`.

cannorin: That would be another PR. We also have this kind of naming in https://github.com/fsprojects/FSharpPlus/blob/master/src/FSharpPlus/Data/NonEmptyList.fs#L75 etc.

PR #396, opened 2020-11-15
Add additional information about CEs

gusty:
This will hopefully add some light about how to use properly lazy/strict CEs

### DIFF
--- a/docsrc/content/computation-expressions.fsx
+++ b/docsrc/content/computation-expressions.fsx
@@ -36,6 +36,21 @@ In other words:
  - ``monad.plus`` Lazy additive monadic builder. Use when you expect one or more results.
  - ``monad.plus'`` is the strict version of ``monad.plus``
 
+Note that a type is either lazy or strict, but it could act as fx or plus at the same time (see below some examples). This means that we need to pay attention when using a CE over a type, if the type is lazy but with use a strict monad, we'll get strict semantics which probably would make no sense, but if we do the opposite we might run into runtime errors, fortunatelly a compile-time warning (or error) will prevent us.
+
+A simple way to find out if a type is strict or lazy is to execute this in fsi: `let _ : MyType<'t> = monad { printfn "I'm strict" }`
+
+For layered monads (monad transformers) the general rule is: the monad is strict unless at least one of its constituent types is lazy, in that case the whole monad becomes lazy.
+
+*)
+
+let _ : OptionT<list<unit option>> = monad { printfn "I'm strict" }
+// will print I'm strict, because OptionT and list are strict
+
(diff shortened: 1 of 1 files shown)


docsrc/content/computation-expressions.fsx
```diff
@@ -36,6 +36,21 @@ In other words:
  - ``monad.plus`` Lazy additive monadic builder. Use when you expect one or more results.
  - ``monad.plus'`` is the strict version of ``monad.plus``
 
+Note that a type is either lazy or strict, but it could act as fx or plus at the same time (see below some examples). This means that we need to pay attention when using a CE over a type, if the type is lazy but with use a strict monad, we'll get strict semantics which probably would make no sense, but if we do the opposite we might run into runtime errors, fortunatelly a compile-time warning (or error) will prevent us.
```
adz, line 39: typo on 'fortunatelly' -> fortunately

adz:
I think it helps!

I'm thinking we should discuss how CE's work, and why it affects the monad builder. However, I haven't got my thoughts together well enough to do so...

gusty:
I agree, the current CE documentation is not the best we can do.

docsrc/content/computation-expressions.fsx
```diff
@@ -31,11 +31,26 @@ There is a single computation expression: ``monad`` but it comes in 4 flavours:
...
 
+Note that a type is either lazy or strict, but it could act as fx or plus at the same time (see below some examples). This means that we need to pay attention when using a CE over a type, if the type is lazy but with use a strict monad, we'll get strict semantics which probably would make no sense, but if we do the opposite we might run into runtime errors, fortunately a compile-time warning (or error) will prevent us.
+
+A simple way to find out if a type is strict or lazy is to execute this in fsi: `let _ : MyType<'t> = monad { printfn "I'm strict" }`
+
+For layered monads (monad transformers) the general rule is: the monad is strict unless at least one of its constituent types is lazy, in that case the whole monad becomes lazy.
+
+*)
+
+let _ : OptionT<list<unit option>> = monad { printfn "I'm strict" }
```
wallymathieu, line 47: I think that you need to import `FSharpPlus.Data`, while also include the reference to the F#+ dll (done in this file a bit further down)

wallymathieu (changes requested):

wallymathieu:
I think the text helps a bit to explain the difference. Nice to have a simple way to test it as well.

docsrc/content/computation-expressions.fsx
gusty: Yes, maybe I can move the open declarations further up, so those one liners compile.

wallymathieu (approved):

PR #404, opened 2020-11-27
Inverted length comparison in String.drop

nicholaslawton:
`String.drop` is a safe version of `String.skip`, which intends to protect a call to `String.skip` with arguments which would cause an exception, and instead return an empty string.

However, the comparison of the length of the input string with the number of characters to drop is inverted. When it would be safe to drop the requested number of characters from the input string, `String.drop` will instead return an empty string; and when the requested number of characters exceeds the length of the input string, `String.drop` will cause the exception from `String.skip`.

The solution is to invert the comparison.


gusty:
Good catch, thanks

Janiczek:
Ah, someone beat me to this :smile: I've just noticed this when solving AdventOfCode ...

PR #407, opened 2020-12-05
Update Task extensions so they can handle exceptions and cancellations

gusty:
Our current implementation is very naïve, for instance it doesn't propagate exceptions properly.

Here we attempt to handle:
 - Exceptions
 - Cancellations
 - Short-circuits


wallymathieu (approved):
Works better than before! 👍

Issue #408, opened 2020-12-08
Option from C# Try method


3Rafal:
When working with C# APIs there are cases when I have an interface

```csharp
public interface IValueGetter
{
    bool TryGetValue(out var myValue);
}
```
So I map it to option like

```fsharp
let optionFromTryGet = function
     | (true, x) -> Some x
     | _ -> None

let optValue = getter.TryGetValue() |> optionFromTryGet
```
I wonder if it can be generalized to Option extension in F#+.
Something like `Option.fromTry` implemented as above helper function.

There are similar implementations already floating around on the internet.
For example this one, which is dedicated to parsing: 
http://www.fssnip.net/2y/title/Functional-wrappers-for-TryParse-APIs

wallymathieu:
See 
https://github.com/fsprojects/FSharpPlus/blob/master/tests/FSharpPlus.Tests/General.fs#L2121-L2136

where you have the customer id from: https://github.com/fsprojects/FSharpPlus/blob/master/tests/CSharpLib/CustomerId.cs#L5

3Rafal:
@wallymathieu , 
I would like to generalize over C# try pattern with out variable. It's used not only for parsing, but also on getting values in general.
See Dictionary example: https://docs.microsoft.com/pl-pl/dotnet/api/system.collections.generic.dictionary-2.trygetvalue?view=net-5.0

This kind of methods return tuple of `bool * 'a` and I would like to be able to convert it to option easily. I've already written this kind of handling a couple of times in my projects, so I guess it could be generalized to a pattern on F# side.

3Rafal:
Also, this pattern was described by Scott Wlaschin here:
https://swlaschin.gitbooks.io/fsharpforfunandprofit/content/posts/completeness-seamless-dotnet-interop.html

He uses couple of different methods with out variable and binds them to `bool * 'a` tuples. But, he doesn't show that you have to pattern match on these tuples to use them. What I would like to do is to pipe them directly to Option monad, so that the handling is easier.

wallymathieu:
I'm unsure if it's possible, but it would be nice to have.

gusty:
Isn't that exactly what our `Option.protect` function does? 

http://fsprojects.github.io/FSharpPlus/reference/fsharpplus-option.html#protect

when combined with an `Option.ofPair` function we could possible add?

3Rafal:
[...]

PR #409, opened 2020-12-09
+ Option.ofPair

3Rafal:
Add function for conversion of C# try pattern to F# option.

### DIFF
--- a/tests/FSharpPlus.Tests/Extensions.fs
+++ b/tests/FSharpPlus.Tests/Extensions.fs
@@ -343,6 +343,16 @@ module Extensions =
     Nullable.toList (Nullable()) |> areEqual []
  
   [<Test>]
+  let ``Option.fromPair returns Some when operation succeeds`` () =
+    Int32.TryParse("123") |> Option.ofPair |> areEqual (Some 123)
+    (dict [("abc",234)]).TryGetValue("abc") |> Option.ofPair |> areEqual (Some 234)
+
+  [<Test>]
+  let ``Option.fromPair returns None when operation fails`` () =
+    Int32.TryParse("abc") |> Option.ofPair |> areEqual None
+    (dict []).TryGetValue("abc") |> Option.ofPair |> areEqual None
+
+  [<Test>]
   let map2Shortest () =
     List.map2Shortest (+) [1;2;3] [2;3] |> areEqual [3;5]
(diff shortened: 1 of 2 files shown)


3Rafal:
This is an implementation of #408 idea

wallymathieu:
Looks like we get:
`System.Xml.XmlException: Name cannot begin with the ''' character, hexadecimal value 0x27. Line 6, position 53` as an error.

tests/FSharpPlus.Tests/Extensions.fs
```diff
@@ -342,6 +342,16 @@ module Extensions =
     Nullable.toList (Nullable 1) |> areEqual [1]
     Nullable.toList (Nullable()) |> areEqual []
  
+  [<Test>]
+  let ``Option.fromPair returns Some when operation succeeds`` () =
```
gusty, line 346: It's `ofPair` not `fromPair` ;)

3Rafal: Changed ;)

3Rafal:
> 
> 
> Looks like we get:
> `System.Xml.XmlException: Name cannot begin with the ''' character, hexadecimal value 0x27. Line 6, position 53` as an error.
> from the XML-documentation.

Fixed :-)

gusty (commented):
This looks OK.

Now the question is does the name of the function makes sense?
Do we all agree?
Or are there better proposals?

@wallymathieu @adz @cannorin @cmeeren

wallymathieu:
I can't come up with a better proposal.

cmeeren:
@gusty Thanks for thinking of me, but I no longer use FSharpPlus due to the IDE performance issues.

gusty:
@cmeeren  Sad to hear that after having you collaborating actively by reporting issues and requesting features for some years.

This function is for Extensions. Does the extension module degrade IDE as well?

Or maybe you don't find Extensions that useful for your projects?

cmeeren:
I haven't considered using only Extensions. Will keep it in mind if I encounter pain points, but I'm doing fine for now. :)

gusty:
Maybe I should find some time again to improve the F# compiler speed for constraint solving.

Thanks one more time for your feedback.

Hopefully is not the last one ;)

wallymathieu:
I think the best bet here is to go with `Option.ofPair` since we don't have any ideas about any better naming.

Issue #386, opened 2020-10-26
Yield! in monad builder


3Rafal:
Currently, F#+ `monad.plus` builder doesn't support `yield!`. 
I think it may be more intuitive to support it, as it would make `monad.plus` similar to FSharp.Core sequence expression.

```fsharp
let l1 = [1;2;3]
let l2 = [4;5;6]

// F# list comp. expr.
let lConcat = [
    yield! l1
    yield! l2 ]

// This doesn't work currently
let lConcat': int list = monad.plus {
    yield! l1
    yield! l2 }

// This works fine
let lConcat'': int list = monad.plus {
    return! l1
    return! l2 }
```

Above example will work correctly if we create `Builder.YieldFrom` method that is an alias to `Builder.ReturnFrom`

gusty:
It would actually be an alias of `return!`. I didn't define it in order to reserve it for future use, but so far I didn't came up with an interesting definition other than that.

I think at this point we could add it as an alias of `return!` for the reasons you exposed.

Does anyone have a strong opinion for/against doing this?

3Rafal:
> It would actually be an alias of return!

@gusty, yes. What I wanted to say is that `Builder.YieldFrom` will have the same implementation as `Builder.ReturnFrom`. That will make `yield!` an alias for `return!`.

3Rafal:
Closed by #410

PR #410, opened 2020-12-13
+ `yield!` to monad.plus

3Rafal:
Added `yield!` operator to monad.plus, so that yield behavior is consistent with return behavior and with FSharp.Core collections expressions.


3Rafal:
This resolves #386

gusty:
I wonder if we should add it to all CEs, not just the plus ones.

For the moment this is safe to go.

Issue #411, opened 2020-12-13
FsCheck tests are not working


3Rafal:
Currently, all of FsCheck tests are false positive.
We use `Check.Quick` method to test properties, but it doesn't throw exception.
If a property fail, it will not be captured by test framework.

The simplest way to wire this properly is using `Check.QuickThrowOnFailure` as shown here:
https://fscheck.github.io/FsCheck/QuickStart.html#Integration-with-xUnit

3Rafal:
I have added a draft proposition (#412) of better FsCheck tests wiring. I've made it a draft, because proper tests fail for multiple cases.

wallymathieu:
At least one of the failures look valid: "get tail from NonEmptyList". It doesn't look like the test make sense for collections of size 1.

wallymathieu:
It also looks as if :
```f#
    let map2Shortest f l1 l2 = { Head = f l1.Head l2.Head; Tail = List.map2 f l1.Tail l2.Tail }
```
should actually be:
```f#
    let map2Shortest f l1 l2 = { Head = f l1.Head l2.Head; Tail = List.map2Shortest f l1.Tail l2.Tail }
```

wallymathieu:
Thanks for the fix @3Rafal ! Thanks for the contribution! 👍

PR #412, opened 2020-12-13
Bugfix: NonEmptyList.map2Shortest, Fix FsCheck tests

3Rafal:
Change FsCheck usage to a version that throws exceptions when properties are not fulfilled.


wallymathieu:
I've added some code to make the tests green. So it looks like you have found a bug 😃 !

3Rafal:
> I've added some code to make the tests green @3Rafal. So it looks like you have found a bug 😃 !

Thanks @wallymathieu ! I think we're fine now 🙂
What do you think @gusty ?

3Rafal:
This resolves #411

wallymathieu:
I've renamed the PR to match what it does. I'm thinking of merging it and putting it on the v1.1 track for release later on today.

gusty (approved):

wallymathieu:
Turns out we have not yet released the map2shortest on the v1.1 track, so it goes into the v1.2 track 😄

Issue #415, opened 2020-12-25
Traverse should stop at the first Error value.


pangwa:
### Description

It should not evaluate list after the first error happend.

### Repro steps

Below code could reproduce the problem:
```f#
let mapper v = 
    Console.WriteLine (sprintf "mapping %d " v)
    if v > 3 then Result.Error(sprintf "error: %d" v)
    else Result.Ok v

let testTrav l = 
    let v = Seq.traverse mapper l 
    v |> Result.map Seq.last

let r = Seq.ofList [1; 2; 3; 4; 5] |> testTrav
match r with
| Result.Ok v -> Console.WriteLine (sprintf "got %d" v)
| Result.Error err -> Console.WriteLine (sprintf "failed %s" err)
```

### Expected behavior
It should prints:

> mapping 1
> mapping 2
> mapping 3
> mapping 4
> failed error: 4

### Actual behavior
The output is:
> mapping 1
> mapping 2
> mapping 3
> mapping 4
> mapping 5
> failed error: 4

### Related information 
Tested with v1.1.6

pangwa:
Closing it as it's been fixed by #416

wallymathieu:
Thanks for finding it @pangwa !

PR #416, opened 2020-12-25
Fix short-circuiting Traverse

gusty:
This PR add short circuit for traversing non-lazy collections and also fixes mapping next element after short-circuiting lazy collections (solves #415), since right now it stops at the first error but maps over the next element after the error, which shouldn't be necessary.


wallymathieu (approved):
👍

PR #417, opened 2020-12-25
Fix finite applicatives traversal tests

3Rafal:
This is mostly a reveral of fixes introduced in: https://github.com/fsprojects/FSharpPlus/commit/75d95689d43da448d3bfc67ed6ffe051648cea9c
There is no failing case, so I assume this could work.

Tests are constructed on implicit assumptions (e.g. magic numbers 4 and 5). Maybe we should improve them?


3Rafal:
@gusty , it seems to be somehow working. I wonder what's your opinion.

gusty (approved):

PR #418, opened 2021-01-07
Evaluate `traverse` left to right

### LINKED ISSUE #414: List traverse runs from right to left
### Description

It looks like List.traverse runs from right to left, it should be from left to right according to the document.

### Repro steps
Below code could reproduce this problem
```fsharp
let mapper (v: int) =
    Console.WriteLine(sprintf "mapping: %d" v)
    Result.Ok v

let test v =
     List.traverse mapper v |> Result.map List.last
   
 test [1; 2; 3]
```
### Expected behavior
It should prints 

> mapping: 1
> mapping: 2
> mapping: 3

### Actual behavior
It prints in the reverse order
> mapping: 3
> mapping: 2
> mapping: 1

### Related information 
The code runs with FSharpPlus 1.1.6

3Rafal:
Fixes #414

### DIFF
--- a/src/FSharpPlus/Control/Traversable.fs
+++ b/src/FSharpPlus/Control/Traversable.fs
@@ -38,6 +38,16 @@ type Traverse =
 
     static member inline Traverse (t: Id<_>, f, [<Optional>]_output: 'R, [<Optional>]_impl: Default3) = Map.Invoke Id.create (f (Id.run t))
 
+    static member inline Traverse (t: _ seq, f, [<Optional>]_output: 'R, [<Optional>]_impl: Default3) =
+       let cons x y = seq {yield x; yield! y}
+       let cons_f x ys = Map.Invoke (cons: 'a->seq<_>->seq<_>) (f x) <*> ys
+       Seq.foldBack cons_f t (result Seq.empty)
+
+    static member inline Traverse (t: _ NonEmptySeq, f, [<Optional>]_output: 'R, [<Optional>]_impl: Default3) =
+       let cons x y = seq {yield x; yield! y}
+       let cons_f x ys = Map.Invoke (cons: 'a->seq<_>->seq<_>) (f x) <*> ys
+       Map.Invoke NonEmptySeq.ofSeq (Seq.foldBack cons_f t (result Seq.empty))
+
     static member inline Traverse (t: seq<'T>, f: 'T->'``Functor<'U>``, [<Optional>]_output: '``Functor<seq<'U>>``, [<Optional>]_impl: Default2) =
         let mapped = Seq.map f t
         Sequence.ForInfiniteSequences (mapped, IsLeftZero.Invoke, List.toSeq) : '``Functor<seq<'U>>``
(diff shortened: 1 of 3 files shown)


3Rafal:
@wallymathieu , it seems that I fail on one Fable test and I don't know why. Is it something obvious that I'm missing?

wallymathieu:
I don't think it's an obvious thing. It looks as if the Fable code assumes that you have a Js iterator and fails due to the fact that what it gets is not something you can iterate over. While debugging locally I get that the value is `1`.

wallymathieu:
Looks like the type inference in this branch differs from master. On master the check is if it's not null or undefined, while on this branch it infers the check to be if it's a non empty sequence.

src/FSharpPlus/Control/Traversable.fs
```diff
@@ -38,16 +38,6 @@ type Traverse =
...
-    static member inline Traverse (t: _ seq, f, [<Optional>]_output: 'R, [<Optional>]_impl: Default3) =
-       let cons x y = seq {yield x; yield! y}
-       let cons_f x ys = Map.Invoke (cons: 'a->seq<_>->seq<_>) (f x) <*> ys
-       Seq.foldBack cons_f t (result Seq.empty)
-
-    static member inline Traverse (t: _ NonEmptySeq, f, [<Optional>]_output: 'R, [<Optional>]_impl: Default3) =
-       let cons x y = seq {yield x; yield! y}
-       let cons_f x ys = Map.Invoke (cons: 'a->seq<_>->seq<_>) (f x) <*> ys
-       Map.Invoke NonEmptySeq.ofSeq (Seq.foldBack cons_f t (result Seq.empty))
-
```
gusty, line 50: Here's thing that might be causing the issue.

Try not changing the overloads and their signature and there should be no reason to change the overload resolution logic in Fable.

If you have to repeat code, go ahead.

gusty (changes requested):

src/FSharpPlus/Control/Traversable.fs
```diff
@@ -95,8 +95,15 @@ type Traverse =
         | Choice2Of2 e -> Return.Invoke (Choice<'U,'Error>.Choice2Of2 e)
 
     static member inline Traverse (t:list<_>   ,f , [<Optional>]_output: 'R, [<Optional>]_impl: Traverse) : 'R =
-       let cons_f x ys = Map.Invoke List.cons (f x) <*> ys
-       List.foldBack cons_f t (result [])
+        let rec loop acc = function
+            | [] -> acc
+            | x::xs -> let v = f x
+                       if v |> IsLeftZero.Invoke |> not
```
gusty: Using `IsLeftZero` here changes the signature and hence the overload resolution logic.
I know it would make it more optimal but optimization should be a separate PR.

gusty (changes requested):

gusty:
Here's a partial explanation of what happened:

- As mentioned, adding the `IsLeftZero` optimization adds an additional constraint
- Calling the `ForInfiniteSequences` creates an additional `Map` constraint, due to the intermediate conversion to list for processing.

Introduction of additional constraints might affect overload resolution logic.

Lesson learned: let's don't try to be too smart in these sensible scenarios. Better to apply small changes, in this case try only to change the desired behavior and leave optimizations for other PRs.

A possible micro optimization here is to change Array.cons definition to something like 

    let cons x (y: _ []) = Array.init (y.Length + 1) (fun i -> if i = 0 then x else y.[i-1])

this will avoid the singleton array allocation, but we have to test if it really impacts execution times. But this is for anothter PR ;)

wallymathieu (approved):
👍

gusty (approved):

PR #433, opened 2021-04-07
+ Internal error to Result.get

gusty:
I am not sure if this is a good design, so did this PR to get started the discussion.

The problem is that, as opposed to `Option.get` where the error doesn't carry any information, in `Result` we normally carry a description of what happened. By using `get` when the value is an `Error` an exception is thrown, hiding the error description.

Of course, we don't know the error type, but in case of an error we can do our best to match on the type and try to throw a better exception.

Comments?

### DIFF
--- a/src/FSharpPlus/Extensions/Result.fs
+++ b/src/FSharpPlus/Extensions/Result.fs
@@ -73,7 +73,7 @@ module Result =
         | Ok x -> x
         | Error e ->
             match box e with
-            | :? string as s -> invalidArg "source" $"Result value was Error: {s}"
+            | :? string as s -> invalidArg "source" ("Result value was Error: " + s)
             | :? exn    as e -> raise <| System.ArgumentException ("Result value was Error", "source", e)
             | _              -> invalidArg "source" $"Result value was Error: {string e}"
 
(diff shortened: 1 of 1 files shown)


adz:
Seems like a good idea to me.

src/FSharpPlus/Extensions/Result.fs
```diff
@@ -68,7 +68,13 @@ module Result =
         with e -> Error e
 
     /// Gets the 'Ok' value. If it's an 'Error' this function will throw an exception.
-    let get (source: Result<'T,'Error>) = match source with Ok x -> x | _ -> invalidArg "source" "Result value was Error"
+    let get (source: Result<'T,'Error>) =
```
wallymathieu, line 71: I don't get why we get the warning around `The type variable 'Error has been constrained to be type 'obj'`

gusty: Oh, that's because the function is not inline, so the use of `string` over a generic type constraint to `obj`.

I think this was somehow fixed in F#5 , for the moment maybe we should either make it inline or use the boxed variable.
Performance is not important here as the application is about to stop at that point.

wallymathieu (approved):
👍

PR #435, opened 2021-04-24
nuget: FSharpPlus

smoothdeveloper:
update the script sample to use nuget reference

### DIFF
--- a/docsrc/content/index.fsx
+++ b/docsrc/content/index.fsx
@@ -54,7 +54,7 @@ Example 1
 This example demonstrates using an extension function defined in this library.
 
 *)
-#r @"../../src/FSharpPlus/bin/Release/net45/FSharpPlus.dll"
+#r @"nuget: FSharpPlus"
 open FSharpPlus
 
 let x = String.replace "old" "new" "Good old days"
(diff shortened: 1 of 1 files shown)


gusty (approved):

docsrc/content/index.fsx
```diff
@@ -54,7 +54,7 @@ Example 1
 This example demonstrates using an extension function defined in this library.
 
 *)
-#r @"../../src/FSharpPlus/bin/Release/net45/FSharpPlus.dll"
+#r @"nuget: FSharpPlus"
```
wallymathieu, line 57: You could do
(**
```f#
#r @"nuget: FSharpPlus"
```
*)
and move the previous reference to the hidden section

smoothdeveloper: My fsdocs foo is subzero, I thought the current change would do what I expect on the website itself.

What is the advantage/outcome of putting it in a markdown codeblock in a comment in your suggestion?

smoothdeveloper: @wallymathieu ok, so if we'd do this, we'd need to add back the reference to the binary in the index.fsx, in a hidden comment, right?

the drawback of keeping it like this is that you can't showcase code you'd have in a branch in that page, so it makes sense I try to adjust it.

I need to pull it and test locally with fsdocs.

smoothdeveloper: ok I think this is good now, I went and tried locally and inspected the html which seems to be as expected.

wallymathieu:
I've published the docs, so we should only wait for it to render on github

PR #446, opened 2021-05-27
Support for %B

Happypig375:
https://github.com/dotnet/fsharp/pull/11603


gusty (approved):
This feature is still not in F# but I saw that it was approved in principle, so I guess it will land at some point. Unless they choose to use a different char for it we should be fine to support it in F#+ scans functions.

Happypig375:
It's been merged into the F# compiler.

PR #448, opened 2021-06-03
Applicative dictionary

gusty:
A sort of "infinite" dictionary would allow an applicative instance.


gusty:
One limitation with this implementation is that it would require the infinite dictionary to be in the first position of the applicative expression, otherwise it would try to iterate through an infinite dictionary.

There are ways to improve this:

 - Create a specific type and match against the type in the applicative implementation. The problem is that the applicative implementation would be linked to only this specific type, though this could be acceptable in order to get applicatives working.
 - Make the applicative implementation to get the count first, then iterate with the smaller dict. This is in anycase a good optimization. The problem here is what if both dictionaries are infinites? And how do we signal that it's infinite? Count=MaxInteger? Or maybe use negative numbers, this feels really hacky but it would allow to encode the infinity nature of the dictionary, moreover if we allow adding elements to these dictionaries we can return a negative count of the overridden elements.

gusty:
This is ready to add either dictionary solution. The error in latest commit is just because a conditional for Fable3 should be added.

There are mainly 2 possible roads:

- Infinite dictionary: Can be anon. Breaks early when trying to iterate. Could be recognized by setting count to -1.
- Defaultable dictionary: Looks like a normal dictionary with all methods, it just return a default value when the key doesn't exists.  It needs a type test in the operation to recognize it.  Doesn't break but it might silently disappear with some operations, like `Dict.map`.

However: trying hard to make `Return` for dictionaries maybe is not worth. Sometimes it's better to have a restriction than fighting it back, ie Semigroups.

Closing this in favor of #575

src/FSharpPlus/Extensions/Dict.fs
```diff
@@ -1,11 +1,176 @@
...
+        member _.Remove (_item: KeyValuePair<'TKey,'TValue>) : bool                     = raise (System.NotImplementedException())
+        *)
+
 /// Additional operations on IDictionary<'Key, 'Value>
 [<RequireQualifiedAccess>]
 module Dict =
     open System.Collections.Generic
     open System.Collections.ObjectModel
 
+    /// <summary>Creates a conceptually infinite dictionay containing the same value for all possible keys.</summary>
```
Happypig375, line 86:
```suggestion
    /// <summary>Creates a conceptually infinite dictionary containing the same value for all possible keys.</summary>
```

PR #452, opened 2021-06-05
Test the type providers

Happypig375:


Happypig375:
The provided types don't seem to be available?

gusty:
@cannorin  maybe you can help here ?

gusty:
On @cannorin  behalf 

Type providers are included in a separate project (FSharpPlus.Providers.fsproj) so it must also be referenced

this is because provided types are erased, so type providers themselves should be placed in a design time dll so that it doesn't get shipped with the resulting executables

the same reason as FSharp.Data has two dlls in their package


so the fix for the test would be just adding one more ProjectReference

gusty (approved):

PR #455, opened 2021-06-24
Add bindError -- flipped `catch` which is useful for piping

adz:
Adding bindError, which I find myself reaching for when piping values, just for a more consistent style.

```
    let subscribe profile watermark =
        subscribeProfileAsResultT watermark profile
        </ catch /> (UserProfiles.Subscribe.deactivateOrResubscribe mitimesUri profile)
        |> ResultT.map ignore
```

Also, I've starting using fantomas to autoformat, which (currently) can't format `</ catch />` nicely. It becomes:
```
    let subscribe profile watermark =
        subscribeProfileAsResultT watermark profile
        </ catch
           /> (UserProfiles.Subscribe.deactivateOrResubscribe mitimesUri profile)
        |> map ignore
```

I also had to recently train a new dev on code and catch and the </ and /> operators were kind of hard to explain.

Having `bindError` as an alternative style, looks like:
```
    let subscribe profile watermark =
        subscribeProfileAsResultT watermark profile
        |> bindError (UserProfiles.Subscribe.deactivateOrResubscribe mitimesUri profile)
        |> map ignore
```

### DIFF
--- a/src/FSharpPlus/Operators.fs
+++ b/src/FSharpPlus/Operators.fs
@@ -982,9 +982,27 @@ module Operators =
     /// <category index="18">Monad Transformers</category>
     let inline throw (error: 'E) : '``'MonadError<'E,'T>`` = Throw.Invoke error
 
-    /// <summary> Executes a handler when the value contained in the Error monad represents an error. </summary>
+    /// <summary>
+    /// Executes a handler when the value contained in the Error monad represents an error.
+    /// This is bindError flipped, which makes it useful when used as an operator.
+    /// </summary>
+    /// <example>
+    /// <code>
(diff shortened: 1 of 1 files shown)


gusty (approved):
It seems reasonable to me.

gusty:
In anycase we should submit an issue to Fantomas.

gusty:
@adz -> `C:\Repos\FSharpPlus\src\FSharpPlus\Operators.fs(985,5): warning FS3390: This XML comment is invalid: 'Name cannot begin with the '|' character, hexadecimal value 0x7C. Line 8, position 38.`

adz:
@gusty those pesky < and >! Have pushed up a separate PR and will merge if it passes. Thanks.

PR #460, opened 2021-07-11
Enable CEs in Fable

gusty:
After testing CEs in Fable I just realized that the problem is the complex resolution of the `Delay` method.

This solution consists in leaving only the default Delay implementation available to Fable, which in theory is less efficient in some cases, but still better than nothing.

### DIFF
--- a/tests/FSharpPlusFable.Tests/FSharpTests/General/Monad.fs
+++ b/tests/FSharpPlusFable.Tests/FSharpTests/General/Monad.fs
@@ -18,7 +18,7 @@ let monad = testList "Monad" [
         equal ["Join"] (SideEffects.get ()))
     #endif
 
-    #if !FABLE_COMPILER
+    #if !FABLE_COMPILER || FABLE_COMPILER_3
     testCase "workFlow" (fun () ->       
         let testVal = 
             monad {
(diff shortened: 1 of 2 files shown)


tests/FSharpPlusFable.Tests/FSharpTests/General/Monad.fs
```diff
@@ -18,31 +23,58 @@ let monad = testList "Monad" [
...
             monad {
                 let! x1 = WrappedListD [1;2]
                 let! x2 = WrappedListD [10;20]
                 return ((+) x1 x2) }
-        Assert.IsInstanceOf<WrappedListD<int>> (testVal))
-
+        #if !FABLE_COMPILER
+        Assert.IsInstanceOf<WrappedListD<int>> (testVal)
+        #endif
+        equal (WrappedListD [11; 21; 12; 22]) testVal
```
wallymathieu, line 36: Isn't this a stronger requirement than the one above?

wallymathieu (approved):
👍

tests/FSharpPlusFable.Tests/FSharpTests/General/Monad.fs
gusty: I just enable the type test and it works, thanks for pointing out.

PR #467, opened 2021-10-05
Fix Seq.drop.

Choc13:
Addresses #468

### DIFF
--- a/src/FSharpPlus/Extensions/Seq.fs
+++ b/src/FSharpPlus/Extensions/Seq.fs
@@ -168,10 +168,12 @@ module Seq =
     ///
     /// <returns>The result sequence.</returns>
     let drop count (source: seq<_>) =
-        let mutable i = count
-        use e = source.GetEnumerator ()
-        while (i > 0 && e.MoveNext ()) do i <- i-1
-        seq { while e.MoveNext () do yield e.Current }
+        seq {
+            let mutable i = count
+
+            for x in source do
+                if i > 0 then i <- i - 1 else yield x
+        }
 
     #if !FABLE_COMPILER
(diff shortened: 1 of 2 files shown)


src/FSharpPlus/Extensions/Seq.fs
```diff
@@ -168,10 +168,12 @@ module Seq =
     ///
     /// <returns>The result sequence.</returns>
     let drop count (source: seq<_>) =
-        let mutable i = count
-        use e = source.GetEnumerator ()
-        while (i > 0 && e.MoveNext ()) do i <- i-1
-        seq { while e.MoveNext () do yield e.Current }
+        seq {
```
Choc13, line 171: If this implementation is less efficient/desirable for some reason (although I don't believe it's creating any additional overhead and it's still lazy) then another solution that is closer to the old code is

```fsharp
let drop count (source: seq<_>) =
    seq {
        let mutable i = count
        use e = source.GetEnumerator ()
        while (i > 0 && e.MoveNext ()) do i <- i-1
        while e.MoveNext () do yield e.Current 
    }
```

As by moving the `seq` around the `use` statement on `GetEnumerator()` it means that it won't be disposed until the `seq` computation has completed.

gusty: Yes, I think that was the original intention in the existing code.

gusty:
Thanks for spotting and fixing this.

Choc13:
No problem, thanks for merging so fast. Any ideas when it will be in a release?

gusty:
We're trying to figure out if we can fix something else and then we'll release

PR #478, opened 2022-04-29
Gusty annotations

gusty:
Add type annotations to unused arguments, hoping it will help Fable type check to workaround a regression in the F# compiler.


gusty:
For the moment this is not needed.

PR #485, opened 2022-07-28
Proper SeqT implementation

gusty:
The previous implementation was not adding any effect besides binding strategy.
With this implementation we can derive both `asyncSeq` as `type AsyncSeq<'T> = SeqT<Async<bool>, 'T>` and `taskSeq` as `type TaskSeq<'T> = SeqT<Task<bool>, 'T>` among combinations with other monads like `type AsyncResultSeq<'T, 'Error> = SeqT<ResultT<Async<Result<bool, 'Error>>, 'T>` .

This adds the interfaces `IEnumeratorM<'Monad<bool>, 'T>` and `IEnumerableM<'Monad<bool>, 'T>` which would ideally be an alias of `SeqT<'Monad<bool>, 'T>` but since as of F# 6 we can't create static members in interfaces we will have to rely in composition instead.

### DIFF
--- a/src/FSharpPlus/Data/Seq.fs
+++ b/src/FSharpPlus/Data/Seq.fs
@@ -92,6 +92,9 @@ type SeqT<'``monad``, 't> =
 
 module SeqT =
 
+    [<Literal>]
+    let private enumNotStarted = "Enumeration has not started. Call MoveNext."
+
     let ofIEnumerableM x : SeqT<'``Monad<bool>``, 'T> = SeqT x
 
     [<RequireQualifiedAccess; EditorBrowsable(EditorBrowsableState.Never)>]
@@ -110,7 +113,7 @@ module SeqT =
                         member _.Current =
                             match current with
                             | Some c -> c
-                            | None -> invalidOp "Enumeration has not started. Call MoveNext."                            
+                            | None -> invalidOp enumNotStarted
                         member x.MoveNext () = monad' {
                             match state with
                             | SeqState.NotStarted inp ->
@@ -144,7 +147,7 @@ module SeqT =
                         member _.Current =
                             match current with
                             | Some c -> c
-                            | None -> invalidOp "Enumeration has not started. Call MoveNext."
+                            | None -> invalidOp enumNotStarted
                         member x.MoveNext () = monad' {
                             match state with
                             | SeqState.NotStarted inp ->
@@ -205,7 +208,7 @@ module SeqT =
                 member _.GetEnumerator () =
                     { new IEnumeratorM<'``Monad<bool>``, 'T> with
                         member _.MoveNext () = result false
-                        member _.Current = invalidOp "Enumeration has not started. Call MoveNext."
+                        member _.Current = invalidOp enumNotStarted
                         member _.Dispose () = () } }
 
     let inline singleton (v: 'T) : SeqT<'``Monad<bool>``, 'T> =
@@ -220,7 +223,7 @@ module SeqT =
                             return res }
                         member _.Current =
                             if started then v
-                            else invalidOp "Enumeration has not started. Call MoveNext."
+                            else invalidOp enumNotStarted
                         member _.Dispose () = () } }
 
     [<EditorBrowsable(EditorBrowsableState.Never)>]
@@ -234,7 +237,7 @@ module SeqT =
                     { new IEnumeratorM<'``Monad<bool>``, 'T> with 
                         member _.Current =
                             match state with
-                            | -1 -> invalidOp "Enumeration has not started. Call MoveNext."
+                            | -1 -> invalidOp enumNotStarted
                             | _  -> current
                         member x.MoveNext () = innerMonad {
                             match state with
@@ -283,7 +286,7 @@ module SeqT =
                         member _.Current =
                             match started with
                             | Some v -> v
-                            | None -> invalidOp "Enumeration has not started. Call MoveNext."
+                            | None -> invalidOp enumNotStarted
                         member _.Dispose () = () } }
 
     [<RequireQualifiedAccess; EditorBrowsable(EditorBrowsableState.Never)>]
@@ -303,7 +306,7 @@ module SeqT =
                         member _.Current =
                             match current with
                             | Some c -> c
-                            | None -> invalidOp "Enumeration has not started. Call MoveNext."                            
+                            | None -> invalidOp enumNotStarted
                         member x.MoveNext () = monad' {
                             match state with
                             | CollectState.NotStarted inp ->
@@ -318,7 +321,7 @@ module SeqT =
                                         let e2 = (f e1.Current :> IEnumerableM<'``Monad<bool>``, 'U>).GetEnumerator ()
                                         state <- CollectState.HaveInnerEnumerator (e1, e2)
                                     else x.Dispose ()
-                                    x.MoveNext () )
+                                    x.MoveNext ())
                             | CollectState.HaveInnerEnumerator (e1, e2) ->
                                 let! (res2: bool) = e2.MoveNext ()
                                 if res2 then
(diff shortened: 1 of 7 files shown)


gusty:
This is almost Ready for review.
Here's what's still pending:

- [ ] Reduce number of constraints
- [ ] Add some docs with sample usage
- [ ] Review naming for `M` functions like `iterM`, `mapM`, `unfoldM` and so on
- [ ] Bring some AsyncSeq tests
- [ ] See if it's worth adding some specific liftable functions for tasks

We won't be adding all AsyncSeq functions as some of them are very specific to Async.

gusty:
Regarding naming:

- `unfold` should be `unfoldM`, then we should add `unfold`.
- `iter`, `iterM` and `mapM` are consistent, although they don't match the generic `iter` and `mapM` but they are not intended to be generalized, Haskell's `ListT` have similar names and they are also not generalized.
- `bindM` in fact is like `lift` them `bind`, therefore renamed to `bindLift` althogh it may remain internal.
- `toSeqM` is `run` that was eveident in the code, so `toListM` and `toArrayM` can be renamed as `runToList` and `runToArray`.

src/FSharpPlus/Data/Seq.fs
```diff
@@ -16,73 +16,885 @@ module Seq =
...
+                            | CollectState.HaveInputEnumerator e1 ->
+                                state := CollectState.Finished
+                                dispose e1
+                            | CollectState.HaveInnerEnumerator (e1, e2) ->
+                                state := CollectState.Finished
+                                dispose e2
+                                dispose e1
+                            | _ -> () } }
+
+    let inline lift2<'T1, 'T2, 'U, .. > (f: 'T1 -> 'T2 -> 'U) (x1: SeqT<'``Monad<bool>``, 'T1>) (x2: SeqT<'``Monad<bool>``, 'T2>) : SeqT<'``Monad<bool>``, 'U> =
```
abelbraaksma, line 399: Out of curiosity, what is `..` in the generic type declaration? Can't believe I've seen that before.

src/FSharpPlus/Data/Seq.fs
```diff
@@ -16,73 +16,885 @@ module Seq =
...
 
-    [<EditorBrowsable(EditorBrowsableState.Never)>]
-    static member inline Lift2 (f: 'T->'U->'V, x: SeqT<'``Monad<seq<'T>``>, y: SeqT<'``Monad<seq<'U>``>) = SeqT.lift2 f x y : SeqT<'``Monad<seq<'V>``>
 
-    [<EditorBrowsable(EditorBrowsableState.Never)>]
-    static member inline Lift3 (f: 'T->'U->'V->'W, x: SeqT<'``Monad<seq<'T>``>, y: SeqT<'``Monad<seq<'U>``>, z: SeqT<'``Monad<seq<'V>``>) = SeqT.lift3 f x y z : SeqT<'``Monad<seq<'W>``>
+    [<RequireQualifiedAccess>]
+    type TryFinallyState<'``Monad<bool>``, 'T> =
+       | NotStarted    of SeqT<'``Monad<bool>``, 'T>
+       | HaveBodyEnumerator of IEnumeratorM<'``Monad<bool>``, 'T>
```
abelbraaksma, line 702: Naming, consider `HasBodyEnumerator` perhaps (same on others)?

src/FSharpPlus/Data/Seq.fs
gusty: That's a rarely used syntax, it means "infer the rest of the type parameters with their constraints". I wish `< .. >` was also allowed.

tests/FSharpPlus.Tests/SeqT.fs
```diff
@@ -0,0 +1,171 @@
...
+        let y = (fn |> SeqT.run |> Reader.run) -1 |> Seq.toList
+        areEqual [] y
+
+
+module ComputationExpressions =
+
+    type __ = bool
+    
+    type Task = Task<__>
+    type Async = Async<__>
```
abelbraaksma, line 115: I haven't seen the other code in depth, so maybe this is a standard pattern, but considering you only use `__` here, is there a difference to not just using `Task<bool>`? Seems clearer that way.

abelbraaksma (commented):
Merely some questions ;). Not yet an in-depth review ;)

src/FSharpPlus/Data/Seq.fs
abelbraaksma: Cool!

tests/FSharpPlus.Tests/SeqT.fs
gusty: Yes. That's unreadable without context.
The context is, I have another PR where I use some kind of consistent Higher Kind encodings, there `__` represents an unapplied type parameter. There `__` is a singleton type, but here it turned out that `bool` works better as it provides some useful information to the Enumerable abstraction. Then I found out that if I consider `__` is `bool`  I have the same notation as in that PR.
I still have to experiment with other monads.

To give you an example, it would read better ``SeqT<Async, 'T>` than `SeqT<Async<bool>, 'T>` because bool is implementation details.

src/FSharpPlus/Data/Seq.fs
gusty: > Merely some questions ;). Not yet an in-depth review ;)

Makes sense, you can't review if things aren't clear.

abelbraaksma:
> _`type TaskSeq<'T> = SeqT<Task<bool>, 'T>`_

I recently experienced quite some issues where `Task` was in an eager sequence. I wonder how that plays out here. Since tasks are hot-started and everything. I solved my issues by wrapping stuff into the `taskSeq` of Don Syme, which is basically wrapping it in an `IAsyncEnumerable`, which I see you're also using here, but only for async sequences. It may be worth testing the task-seq code with significantly large sequences of tasks in lists, for instance (which is where I had the issue).

I should probably go over your code in a little bit more detail, it's certainly very interesting!

> _toSeqM is run that was eveident in the code, so toListM and toArrayM can be renamed as runToList and runToArray._

not sure if the terminology of `run` would be clear here? `toXXX` and `ofXXX` seem to be the more common naming conventions for conversion from/to types.

gusty:
> I wonder how that plays out here. Since tasks are hot-started and everything. I solved my issues by wrapping stuff into the `taskSeq` of Don Syme, which is basically wrapping it in an `IAsyncEnumerable`, which I see you're also using here,

Not really, I'm using a generic `IEnumerableM<_, _>` but when specialized it behaves in a similar way to `IAsyncEnumerable<_>`, then there are some functions to convert back and forth to `IAsyncEnumerable<_>`.

> but only for async sequences. 

Nope, it's generic so it works with both async and tasks.

> It may be worth testing the task-seq code with significantly large sequences of tasks in lists, for instance (which is where I had the issue).

Absolutely, but I think both implementations relies on the fact that a strict monad (task) combined with a lazy one (seq) results in a lazy combination.

> I should probably go over your code in a little bit more detail, it's certainly very interesting!

Sure, thanks !

> not sure if the terminology of run would be clear here? toXXX and ofXXX seem to be the more common naming conventions for conversion from/to types.

Well that's a bit tricky I know. The context is that the `run` functions are standard in this library and they all mean `unwrap` the monad transformer, so for instance if you have `async` with `option`, you wrap them in `OptionT`, when you `run` it you get back the `async<option<_>>`, here in `SeqT` you get `async<seq <_>>`, and the other functions are like running `run` and the mapping `toX` collection.

In the original asyncSeq library these functions are not generic so naming them is easier, they are called `toSeqAsync` and so one, here initially I name them `toSeqM` by applying the initial renaming pattern of `async` to `M`, but I don't think those name are clear enough either. I'm open to suggestions.

wallymathieu (approved):

PR #488, opened 2022-08-28
Use FSharp.Collections.List instead

wallymathieu:


src/FSharpPlus/Extensions/List.fs
```diff
@@ -177,18 +173,12 @@ module List =
     /// <param name="lst">The input list</param>
     /// 
     /// <returns>For invalid indexes, the input list.  Otherwise, a new list with the item removed.</returns>
-    let removeAt i lst =
-        if List.length lst > i then
-            lst.[0..i-1] @ lst.[i+1..]
-        else lst
+    let removeAt i lst = FSharp.Collections.List.removeAt i lst
```
gusty, line 176: > For invalid indexes, the input list.

No, the FSharp.core version throws.

gusty (changes requested):

src/FSharpPlus/Extensions/List.fs
wallymathieu: So should we keep this implementation and use an obsolete attribute

gusty: IMHO: No, yes.

No, we need to rename it to something else (first think that comes to my mind is `deleteAt`) and 
yes, issue a deprecate warning, saying something like "This function was included in FSharp.Core but throwing,  use `deletaAt` (or whichever name we choose) instead ot if you want to throw exceptions use the full path to `removeAt` in Core ".

src/FSharpPlus/Extensions/List.fs
```diff
@@ -47,9 +47,86 @@ module List =
...
+    ///
+    /// <example id="skip-3">
+    /// <code lang="fsharp">
+    /// let inputs = ["a"; "b"; "c"; "d"]
+    ///
+    /// inputs |> List.skip -1
+    /// </code>
+    /// Evaluates to <c>["a"; "b"; "c"; "d"]</c>.
+    /// </example>
+    let skip count list = FSharp.Collections.List.skip count list
```
gusty, line 129: Let's add same type annotations as FSharp.Core in order to match it 100%, and also because it's a good practice in anycase.

wallymathieu: What type annotations do mean?

gusty: image

vs

image

gusty: also change `i` to `count`

wallymathieu: Those are mainly the xml docs so should already be added

wallymathieu: Or do you mean the `'T` vs `'a` ?

gusty: Yes, for that you have to type annotate the params.

wallymathieu:
What worries me is that the tests are taking significantly longer to complete on this branch.

wallymathieu:
I don't seem to be able to find Seq.removeAt @gusty ?

gusty:
Yes, probably doesn't exist.

src/FSharpPlus/Extensions/List.fs
```diff
@@ -73,15 +92,96 @@ module List =
     #endif
 
     /// Returns a list with all possible tails of the source list.
-    let tails x = let rec loop = function [] -> [] | _::xs as s -> s::(loop xs) in loop x
+    let tails list  = let rec loop = function [] -> [] | _::xs as s -> s::(loop xs) in loop list : list<list<'T>>
```
gusty, line 95: Doble space here
```suggestion
    let tails list = let rec loop = function [] -> [] | _::xs as s -> s::(loop xs) in loop list : list<list<'T>>
```

src/FSharpPlus/Extensions/List.fs
```diff
@@ -258,18 +358,29 @@ module List =
...
+         if List.length lst > i then
+             lst.[0..i-1] @ lst.[i+1..]
+         else lst
+
+    /// <summary>Attempts to remove an item from a list.</summary>
+    /// <param name="i">The index of the item to remove </param>
+    /// <param name="lst">The input list</param>
+    ///
+    /// <returns>For invalid indexes, the input list.  Otherwise, a new list with the item removed.</returns>
+    /// <remarks>Use deletaAt instead or if you want to throw exceptions use the full path to removeAt in Core until this function is removed from this library.</remarks>
```
gusty, line 371:
```suggestion
    /// <remarks>Use deletaAt instead or if you want to throw exceptions use the full path to removeAt in FSharp.Core until this function is removed from this library.</remarks>
```

gusty (approved):
Looks good, good job.
Small text adjustments, would it be possible to replace FSharpCore with FSharp.Core or is there any issue in having the `.` inside XML comments?

Also overrides sounds worst than what actually is, maybe it's better to say "it's being redirected to", or is an alias of.

wallymathieu:
I think it's fine to simply mention that we will remove the F#+ version in the next major release.

PR #489, opened 2022-08-31
Add nuget reference to each doc and upgrade to fsdocs-tool

wallymathieu:
Remove the need for .net core 3
Add F#+ nuget reference in the docs
```f#
 #r @"nuget: FSharpPlus"
 ```
See for instance https://github.com/fsprojects/FSharpPlus/blob/db2ce202e3ee2b305f755bd5fed233b2b3d0635a/docsrc/content/abstraction-alternative.fsx#L80-L84 for an example


gusty (approved):
Looks good but let's change the name of the PR

PR #500, opened 2022-09-07
-.net framework

wallymathieu:
In order to speed up builds and to allow development environment to be easy to start with it would be nice to drop .net framework from the builds since the later versions of .net framework can consume .net standard. 

https://docs.microsoft.com/en-us/dotnet/standard/net-standard?tabs=net-standard-2-0


wallymathieu:
It's merged

PR #501, opened 2022-09-07
Make left zero dynamic

gusty:
The goal is to speed up and ease type inference with a minimum (one off) runtime cost.


gusty:
Closing in favor of #503

PR #507, opened 2022-10-03
Speed up Array extensions

fcallejon:


src/FSharpPlus/Extensions/Array.fs
```diff
@@ -36,22 +37,88 @@ module Array =
...
+        source
+        |> Array.toSeq
+        |> Seq.replace oldValue newValue
+        |> Seq.toArray: 'T []
+    #else
+        match source with
+        | null -> null
+        | [||] -> [||]
+        | _ ->
+            let candidate = new ResizeArray<'T>()
```
gusty, line 82: Try `ArrayCollector` instead, I think it's faster.

gusty (changes requested):
Excellent !
Looks good to me, just a few details.

gusty (approved):
Looks good, thanks !

PR #512, opened 2022-11-12
Raise ArgumentNullException for Array functions

fcallejon:


src/FSharpPlus/Extensions/Array.fs
```diff
@@ -6,6 +6,13 @@ module Array =
 
     open System
     open FSharp.Core.CompilerServices
+    
+    let inline private raiseIfNull paramName (paramValue: _ seq) =
```
gusty, line 10: Instead of `_ seq` I would leave it generic.
Probably you did it like that in order to make it work for separators which is a seq of arrays.
It will null check the sequence, but it won't null check every single element in the sequence, which is kind of ok, since the null check will occur downstream in any case.

gusty (changes requested):
Thanks, just a few adjustments and we merge it.

src/FSharpPlus/Extensions/Array.fs
gusty: Also, consider moving this function to Internals.fs you can put it in either the Prelude or the Errors module. This way we can use it for other modules like Seq.

fcallejon: Perfect, will do so.
I wasn't sure if we wanted this to be reused.

gusty (approved):

PR #513, opened 2022-11-13
V1.3 F# core 6

wallymathieu:


src/FSharpPlus/Extensions/Array.fs
```diff
@@ -36,22 +37,89 @@ module Array =
...
     /// Creates a sequence of arrays by splitting the source array on any of the given separators.
     let split (separators: seq<_ []>) (source: _ []) = source |> Array.toSeq |> Seq.split separators |> Seq.map Seq.toArray
 
     /// Replaces a subsequence of the source array with the given replacement array.
-    let replace oldValue newValue source = source |> Array.toSeq |> Seq.replace oldValue newValue |> Seq.toArray                : 'T []
+    let replace (oldValue: 'T seq) (newValue: 'T seq) (source: 'T[]) : 'T[] =
+    #if FABLE_COMPILER || NET45
+        source |> Array.toSeq |> Seq.replace oldValue newValue |> Seq.toArray: 'T []
+    #else
+        let oldValueArray = oldValue |> Seq.toArray
```
wallymathieu, line 78: I had to change to toArray here since the actual type signature in the released version of F#+ is seq as can be seen for instance here:
https://www.fuget.org/packages/FSharpPlus/1.2.5/lib/netstandard2.0/FSharpPlus.dll/FSharpPlus/Array
@fcallejon

src/FSharpPlus/Data/Validation.fs
```diff
@@ -228,13 +229,20 @@ module Validation =
     /// A tuple with both resulting lists, Success are in the first list.
     /// </returns>
     let partition (source: list<Validation<'TErrors, 'T>>) =
+    #if FABLE_COMPILER || NET45
```
wallymathieu, line 232: Note that both net45 and Fable lack ListCollector @gusty

PR #518, opened 2022-11-14
Added computation expression for validation and added some extensions methods

rodriguestiago0:


src/FSharpPlus/Data/Validation.fs
```diff
@@ -243,6 +243,29 @@ module Validation =
...
+                | Success _, Failure r -> Failure r
+                | Failure r, Failure l -> List.append r l |> Failure
+
+            member _.Return(x) =
+                Success x
+                
+            member _.ReturnFrom(x) =
+                x
+
+        let validator = ValidationBuilder()
```
gusty, line 268: Note that this is just a special case of the `applicative` computation expression, closed to `Validation<list<string>, 'T>`.

I mean, right now you can just define it like: `let validator<'T> = applicative<Validation<list<string>, 'T>>` and you'll have the same methods and more available.

src/FSharpPlus/Extensions/Validation.fs
```diff
@@ -0,0 +1,80 @@
...
+    
+    let inline validate errorMessage f v =
+        if f v then
+            Success v
+        else
+            Failure [ errorMessage ]
+
+    let inline requireString propName =
+        let errorMessage =
+            sprintf "%s cannot be null, empty or whitespace." propName
```
gusty, line 16: In my experience using string as errors doesn't scale well, unless they are at the boundaries with the user.
But otherwise, validations errors should be represented with a custom type, normally a DU, which should be directly related to the business model of validation errors.

Having said that, we could add an additional parameter which is a function that creates the error, so the user is free to supply whatever structure he uses.

src/FSharpPlus/Data/Validation.fs
gusty: Actually the `bind` is not part of the `applicative` CE, but `Validation` is not a monad, as it doesn't obey monad rules. However in order to get the short-circuit behavior you can use `let validator<'T> = monad'<Result<'T, List<string>>`

src/FSharpPlus/Extensions/Validation.fs
xperiandri: @rodriguestiago0 
1. Error strings must be declared as public static functions to be available for plain reuse (for example in unit tests)
2. We have a case right now when we need to delay error string creation till it is required:
    a. Output to log as is
    b. Output to API response with camel case field name modification

src/FSharpPlus/Data/Validation.fs
xperiandri: Maybe an alias then?

xperiandri (changes requested):

gusty:
> Maybe an alias then?

That's an old discussion we have here.
This library contains highly reusable code
We can create a module with some CEs, but as you can see defining aliases are one-liner.

My original view was that F#+ is kind of a library for libraries, meaning that you can have your own enterprise specific libraries where you take what you need, the way you need, this is giving a specific name and closing it over certain type parameters.

I had feedback that some users did that at the beginning, which lead to the desire of making CEs more parameterizable, we did that in v1.2

However, if everybody is strongly for providing aliases we can evaluate the impact. Maybe provide a script with those aliases or a separate package.

src/FSharpPlus/Extensions/Validation.fs
rodriguestiago0: I know to let the user define the Error to be added. It can be a string, int or whatever the developer wants.

tests/FSharpPlus.Tests/Validations.fs
```diff
@@ -11,7 +11,27 @@ module Validation =
...
+  
+  let private isFailure =
+    function
+    | Success _ -> false
+    | Failure _ -> true
+    
+  let private getSuccess =
+    function
+    | Success s -> s
+    | Failure _ -> failwith "It's a failure"
```
gusty, line 28: We have `defaultValue` and `defaultWith`, same as in `Result` which now are part of FSharp.Core.

I think having a function throwing a fixed exception, is not worth, we can just use the above functions in a more explicit way. It's also more predictable from the user perspective.

tests/FSharpPlus.Tests/Validations.fs
```diff
@@ -11,7 +11,27 @@ module Validation =
...
+  
+  let private isSuccess =
+    function
+    | Success _ -> true
+    | Failure _ -> false
+  
+  let private isFailure =
+    function
+    | Success _ -> false
+    | Failure _ -> true
```
gusty, line 23: These can be considered for addition.

src/FSharpPlus/Extensions/Validation.fs
```diff
@@ -0,0 +1,65 @@
+﻿namespace FSharpPlus
+
+[<RequireQualifiedAccess>]
+module Validations =
```
gusty, line 4: What about `Validators` ?

gusty: Or maybe `Require` inside `Validation` ? Then we shorten all names.

tests/FSharpPlus.Tests/Validations.fs
letstrodrigues: what does that mean?

gusty: I mean we can add them to `Validation`

tests/FSharpPlus.Tests/Validations.fs
rodriguestiago0: both already exist for the Validation type

src/FSharpPlus/Extensions/Validators.fs
```diff
@@ -0,0 +1,75 @@
+﻿namespace FSharpPlus
+
+[<RequireQualifiedAccess>]
+module RequiredValidation =
+    open System
+    open FSharpPlus.Data
+    
+    let inline validate error f v =
```
gusty, line 8: We have this function already and it's being deprecated.

src/FSharpPlus/Extensions/Validators.fs
```diff
@@ -0,0 +1,75 @@
...
+        if f v then
+            Success v
+        else
+            Failure [ error ]
+
+    let inline string error value =
```
wallymathieu, line 14: I'm unsure if just calling it `string` makes it obvious. Would it be better to call it `notNullOrWhiteSpace` ?

src/FSharpPlus/Extensions/Validators.fs
```diff
@@ -0,0 +1,75 @@
...
+        let check (v: string) =
+            try
+                let _ = Net.Mail.MailAddress(v)
+                true
+            with
+            | ex -> false
+
+        validate (error value) check value
+
+    let inline guid error value =
```
wallymathieu, line 33: this type of validator goes into the question if there is some more general class of nonEmpty restrictions?

wallymathieu:
There seems to be prior work that we can reuse some of the thoughts from. For dynamic languages we have:

https://clojure.org/about/spec
https://github.com/jquense/yup

While for F# we have

https://github.com/lfr/FSharp.Domain.Validation
https://github.com/JamesRandall/AccidentalFish.FSharp.Validation

and the new library:

https://github.com/pimbrouwers/Validus

Perhaps we can reuse some of the design work while being more in line with the vision that you promote in this PR (possible to write strictly typed validations instead of stringly typed validation keys/messages).





[6:37](https://fsharp.slack.com/archives/D901ETYRG/p1668962234989539)
https://github.com/jquense/yup

wallymathieu:
Some of the operators seems to not have stabilised in Validus yet. The different libraries all have their built in validators that we can look at. Seems like there is some polymorphism going on for some of the validators.

PR #523, opened 2022-11-23
Add support for ValueTask

rodriguestiago0:


gusty (commented):
Look good so far.
Please update the docs and we're done !

wallymathieu:
I've split the net7 upgrade into #525

abelbraaksma (commented):
There’s a line has a check warning. Is that new with net7? 

```
Check warning on line 43 in src/FSharpPlus/Control/Monoid.fs

 GitHub Actions
/ testFable3SubsetOnCore

This upcast is unnecessary - the types are identical
```

Also:

> Possible incorrect indentation: this token is offside of context started at position (234:1).

wallymathieu:
The issue with fable tests not completing seems to be related to having net7 installed. It works fine if you only have net6.

gusty:
> And it appears that Fable3 tests don’t finish. @gusty is that common?

No @abelbraaksma is not, I will report this to the F# team, it might be the mysterious bug they are hunting for, where the compilation never stops.

In the meantime, the workaround is here: #525

PR #525, opened 2022-11-26
Upgrade tests and build to net7 except for Fable (net6)

wallymathieu:
Co-authored-by: @rodriguestiago0


tests/FSharpPlusFable.Tests/FSharpPlusFable.Tests.fsproj
```diff
@@ -1,49 +1,49 @@
...
-  <PropertyGroup>
-    <OutputType>Exe</OutputType>
-    <Configurations>Debug;Release;Fable;Fable3</Configurations>
-    <Platforms>AnyCPU</Platforms>
-    <LangVersion Condition=" '$(Configuration)' == 'Fable' OR '$(Configuration)' == 'Fable3' ">6.0</LangVersion>
-    <DefineConstants Condition=" '$(Configuration)' == 'Fable'">$(DefineConstants);FABLE_COMPILER;FABLE_COMPILER_FAKE</DefineConstants>
-    <DefineConstants Condition=" '$(Configuration)' == 'Fable3'">$(DefineConstants);FABLE_COMPILER;FABLE_COMPILER_3;FABLE_COMPILER_FAKE</DefineConstants>
-    <TargetFramework>net6.0</TargetFramework>
-  </PropertyGroup>
+	<PropertyGroup>
```
abelbraaksma, line 3: Why the 8 space indent? I think 2 or 4 are common. Probably best to have formatting changes in a separate PR and doing it across the board, though in this case, I think the default of 2 spaces was fine.

abelbraaksma (changes requested):
There are some warnings, where they here before? See Files tab. For instance:

> Possible incorrect indentation: this token is offside of context started at position (234:1). Try indenting this token further or using standard formatting conventions.

abelbraaksma:
Brew, just curious, but looking at the changes, this doesn’t create a mandatory dep on net7 for the libs, right? I mean, I’d assume for the time being that ‘lowest denominator’ for nuget packages will be the rule, unless that’s no longer possible.

wallymathieu:
Notice that the library still has netstandard as the target framework. The build runs on net7 though.

tests/FSharpPlusFable.Tests/FSharpPlusFable.Tests.fsproj
wallymathieu: It could be when Visual Studio does changes

abelbraaksma:
got it ☝️ , thanks.

tests/FSharpPlusFable.Tests/FSharpPlusFable.Tests.fsproj
wallymathieu: looks like it was spaces but has been changed to tabs.

abelbraaksma: VS will honor the `.editorconfig` settings. However, maybe this project does not have specific settings for `fsproj` files. In which case it will use whatever default your system has.

Here's a snippet for `.editorconfig`, maybe try that and re-edit in your editor, see what happens?

```
# Project files and app specific XML files
[*.{csproj,fsproj,shproj,projitems,props,targets,xaml}]
trim_trailing_whitespace = true
insert_final_newline     = true
indent_style             = space
indent_size              = 2
charset                  = utf-8
```

wallymathieu: I agree. We should have that setting in editorconfig. Looks like we have not added it yet for this repository. Perhaps something you want to make a PR for?

PR #547, opened 2023-04-07
Option.bindError

gusty:
I propose to add this function which would allow to attempt to recover from an error condition.
This is similar to our generic `catch`, or the non generic `Result.bindError` but taking always `()` as input.


gusty:
This exists already in FSharp.Core as `Option.orElseWith`

PR #549, opened 2023-09-15
Add curry/uncurry with flip

gusty:
This will add overloads for curryFlip and uncurryFlip


smoothdeveloper:
https://github.com/fsharp/fslang-suggestions/issues/1039#issuecomment-1721528426

> @gusty, do you know if it is possible to make a function that is like flip that would work on any method, and puts the first argument as the last and untuple the arguments? It could be used to turn the extension methods into functions that you can pipe into, passing the this last.

PR #550, opened 2023-09-16
+ DateOnly as Monoid and Zero for TimeOnly

gusty:
The idea is to have some nice properties enabled, as long as they make sense.


wallymathieu (approved):

wallymathieu:
This is solidly in the F#+ 2 track. Would be neat to have it though 😃

PR #552, opened 2023-09-16
NonEmptySeq with static members

gusty:
As from F# 8 we will be able to add static members to interfaces.

In theory, we no longer need to declare our interfaces upfront and add their static members in each Invokable, which creates an undesired (internal for the moment) dependency and also pollutes the Invokable overloads, which has the negative effect of slowing down overload resolution.


gusty:
Continued in #631

PR #554, opened 2023-09-19
Add a function for creating a Task from a value

gusty:
We normally uses `Task.FromValue` but nowadays, using the `task` CE the namespace `System.Threading.Tasks` is not always opened and opening it only to return a task doesn't feel so good.

Sometimes what I do is I use `result` but in some places is too generic.

Another option is writing `task { return x }` but I think is overkill.

Given that we have our Task module in scope with many functions, maybe natural way is to add one there like `List.singleton` does.

### DIFF
--- a/src/FSharpPlus/Extensions/Task.fs
+++ b/src/FSharpPlus/Extensions/Task.fs
@@ -285,6 +285,9 @@ module Task =
             (fun () -> body disp)
             (fun () -> if not (isNull (box disp)) then disp.Dispose ())
 
+    /// Creates a Task from a value
+    let ofValue value = Task.FromResult value
+    
     /// Raises an exception in the Task
     let raise (e: exn) =
         let tcs = TaskCompletionSource<'U> ()


Happypig375:
Why not `Task.singleton`?

gusty:
Thanks, the idea is to get some feedback about names first, before merging this.
If you ask me, it looks like `singleton` is used in list-like collections, as opposed to having multiple values, you have a singleton one.

abelbraaksma:
I agree that this function should be added, maybe also for ValueTask and Async? I think that @Happypig375 refers to the use of `Task.singleton` and the like from FsToolkit (see https://github.com/search?q=repo%3Ademystifyfp%2FFsToolkit.ErrorHandling%20singleton&type=code).

But it feels wrong, 'singleton' suggests that there can also be more. I've seen, in the wild,  Task.ret`, `Task.retn` which may be a good alias, but I think most people would naturally search for something that starts with `of`, like `Task.ofAsync`, `Task.ofUnit` (from non generic tasks), `Task.ofValueTask` etc. 

But keep in mind that there was an effort underway where we wanted a very thin library for these super trivial extensions for F#, because these trivial ones seem to be created and exposed in a bunch of libraries, often with different names. 

The thin lib could be source-embedded. I forgot the most recent state of that discussion, I'll look it up.

gusty:
I like `Task.result` or `Task.ofResult` as it matches the generic `result` function and also the `.FromResult` BCL method.

The problem is, what if in the future we decide to add an `.ofResult` which takes a Result value and make it a Task with either a value or an exception (failed task) ?

Keeping that in mind I still consider `Task.result` (without 'of') as a valid candidate.

gusty:
This PR was already merged my mistake.

I'm feeling that `Task.result` should be the name of the function, for reasons explained above.

Against `.ofValue` I have to say that there is no precedent of such name here nor  in FSharp.Core.

Is anyone strongly opposed to `Task.result`? If so please shout.

gusty:
Both `Task.result` and `ValueTask.result` added at [aae4533](https://github.com/fsprojects/FSharpPlus/commit/aae45336213686660426f5f0a604984b4d6e73e3)

PR #557, opened 2023-09-30
Trace usage

wallymathieu:
- Test traversal overload during tests


gusty (approved):
This looks really nice. We should have done it long time ago.
There is a huge diff in the  .sln file, is that really needed?

wallymathieu:
> This looks really nice. We should have done it long time ago. There is a huge diff in the .sln file, is that really needed?

The main reason for the huge diff is due to the fact that Visual Studio configures the solution to specify that the build should use Test-configuration in some cases.

wallymathieu:
Perhaps it could be that Visual Studio is confused around the identity of the FSharpPlus fsproj and has recreated the configurations.

PR #558, opened 2023-10-02
Adjust parse / tryParse

gusty:
Functions `parse`  and `tryParse` should behave the same, the only difference should be in throwing or returning `None`.

### DIFF
--- a/src/FSharpPlus/Control/Converter.fs
+++ b/src/FSharpPlus/Control/Converter.fs
@@ -165,8 +165,9 @@ type Parse =
 
     #if !FABLE_COMPILER
     static member Parse (_: DateTime      , _: Parse) = fun (x:string) ->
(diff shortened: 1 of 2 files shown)


src/FSharpPlus/Control/Converter.fs
```diff
@@ -139,6 +163,15 @@ type Parse =
             | _         -> invalidArg "value" ("Requested value '" + x + "' was not found.")
         ) : 'enum
 
+    #if !FABLE_COMPILER
+    static member Parse (_: DateTime      , _: Parse) = fun (x:string) ->
+        try DateTime.ParseExact       (x, [|"yyyy-MM-ddTHH:mm:ss.fffZ"; "yyyy-MM-ddTHH:mm:ssZ"|], null, DateTimeStyles.RoundtripKind)
```
wallymathieu, line 168: Could the TryParse implementation be used instead?

wallymathieu (approved):
I think it looks better in the TryParse case.

wallymathieu (approved):

Issue #280, opened 2020-02-25
Validation as a ZipFunctor?


NickDarvey:
I'm not even sure if I'm phrasing this right, but, is `Validation` a `ZipFunctor`? And would this be a valid implementation of it?

```fsharp
module Validation =
  let zip = liftA2 tuple2
```

I can open a PR if it actually is a zippable functor.

gusty:
That's a very interesting question.

Unfortunately I think the answer is No and I'll tell you why:

Let's suppose it is a ZipFunctor which means that it has a `zip` operation which is the inverse of Functor's `unzip` (which is defined as `unzip s = map fst s, map snd s`).

Then we can do:

```
> zip (Validation<string,int>.Success 42) (Validation<string,float>.Failure "Error") ;;
val it : Validation<string,(int * float)> = Failure "Error"
```

and now

```
> unzip it ;;
val it : Validation<string,int> * Validation<string,float> =
  (Failure "Error", Failure "Error")
```
but now

```
> it ||> zip;;
val it : Validation<string,(int * float)> = Failure "ErrorError"
```
so we came back to a different value !

Note however that `Result` seems to be well behaved in this regard:

```
> zip (Result<int,string>.Ok 42) (Result<float,string>.Error "No") ;;
[<Struct>]
val it : Result<(int * float),string> = Error "No"

> unzip it ;;
val it : Result<int,string> * Result<float,string> = (Error "No", Error "No")

> it ||> zip;;
[<Struct>]
val it : Result<(int * float),string> = Error "No"
```

Does it makes sense?

NickDarvey:
That does make sense. Thank you for the accessible explanation @gusty 😊

PR #559, opened 2023-10-12
Non sequential Applicatives

### LINKED ISSUE #36: Traverse asyncs in parallel?
Should ``sequenceA [|x;y;z|]`` (for x, y and z asyncs) be semantically equivalent to ``Async.Parallel [|x;y;z|]`` ?

The operations are equivalent and currently will yield the same results, but ``sequenceA`` (or ``traverse id``) will execute each async in sequence. Well that's probably what the name means ;)

The same way it is possible to implement a specific instance for ``zip`` where it zips two asyncs in parallel, but the question is: is it the right thing to do?

Any thoughts?

gusty:
This is a key feature, we introduce a "parallel" instance of Applicatives which are normally not monads and therefore don't have sequential behavior.

The goal is to remove a bit friction when using the "interesting" applicative instances without breaking some basic rules which provides useful properties, but adding a different set of operation instead of the classical solution of new-type which involves wrapping and unwrapping all the time and pollutes a lot a simple piece of code.

We call them ZipApplicatives as existing F# core zip-like operations like `zip` tend to match them.

It's hard to define additional rules for them, but I would say in general we can expect the full `zip` -`unzip` roundtrip.
Nevertheless the rule that they don't obey is sequential guarantee, as opposed to the ones derived by Monads.

For Validation it was stated the above roundtrip doesn't hold (see #280), but actually if the monoid type is a set-like it does.

In general, list like will behave point-wise and computations will start in sequence but can finish in any order. The latter is not exactly the same as full parallel which would require using the threadpool, see https://github.com/dotnet/fsharp/issues/10301#issuecomment-734280586 for more context on this distinction.

This design takes into consideration the principle of less surprising, namely:

 - Not breaking the rule of `<*>` equivalent to `</ap/>` which would give bad surprises in refactorings/optimizations.
 - Attempt to keep as much as possible existing F# function names, and introducing new ones only in cases where it would otherwise contradict the expectation.

```fsharp
#r @"C:\Repos\FSharpPlus\src\FSharpPlus\bin\Release\netstandard2.1\FSharpPlus.dll"

open FSharpPlus

let r0 = (+) <!> [500; 1000] <.> [1;2;3]
// val r0: int list = [501; 1002]

let x1 = Ok 1
let y1 = Error ["2 didn't work"]
let z1 = Error ["3 didn't work"]

let r1 = app {
    let! x = x1
    and! y = y1
    and! z = z1
    return (x + y + z)
}

// val r1: Result<int,string list> = Error ["2 didn't work"; "3 didn't work"]


let y2 = [2;20;30]
let z2 = [5;50;500; 100000000]

let r2 = app {
    let! x = app { return 1000 }
    and! y = y2
    and! z = z2
    return (x + y + z)
}

// val r2: int list = [1007; 1070; 1530]
```
Fixes #36 
Partially inspired by [Parallel](https://pursuit.purescript.org/packages/purescript-parallel/7.0.0/docs/Control.Parallel)


src/FSharpPlus/Extensions/Task.fs
```diff
@@ -45,7 +45,7 @@ module Task =
     /// <param name="f">The mapping function.</param>
     /// <param name="x">First task workflow.</param>
     /// <param name="y">Second task workflow.</param>
-    let map2 (f: 'T -> 'U -> 'V) (x: Task<'T>) (y: Task<'U>) : Task<'V> =
+    let lift2 (f: 'T -> 'U -> 'V) (x: Task<'T>) (y: Task<'U>) : Task<'V> =
```
wallymathieu, line 48: This looks like a v2 change

gusty: > This looks like a v2 change

`map2` is still available below.

wallymathieu (approved):

PR #569, opened 2023-11-22
Remove subsumption from Applicatives

gusty:
Same as #207 but rebased.

I gave it another try and looks like I found a way not to break anything, so no need to wait for v2.

This will allow Vector and Matrix to implement `IEnumerable`.
Another reason to implement this is removing dependencies between `NonEmptySeq<'T>` (now that F#8 allows static member on interfaces) and `FSharpPlus.Control` by creating its own static members #552 

Also to the original issues:

> Right now there are 2 issues ..

>  - Types implementing `IEnumerable<_>` would default to the `seq<_>` instance. This might be arguably desired.

>  - Types implementing `IEnumerable<_>` would default to the `seq<_>` instance, even if they implement their own applicative > instance. This is not acceptable.

.. I will amend the former like this:

  - Types implementing `IEnumerable<_>` would default to the seq<_> instance. This might be arguably desired, but it will result in unexpected behavior: different IEnumerables can have different applicative instances, this is becoming more evident in #559  for example a ZipList can be an IEnumerable but its applicative instance won't have the same result as if you cast it to seq and then use the applicative for sequences, therefore it's better to require either an explicit applicative implementation, or an explicit cast to `seq`. The other issue is the resulting type will be a `seq<_>` instead of the original type, which would break many things.

**TODO**

- [x] Add tests for the new functionality
- [x] Rebase on #570 just to be sure nothing breaks
- [x] Reconsider final shape of the overloads (tuple, ref, struct tuple)
- [x] Reconsider the dummy type KeyValuePair2, move it to internals, use instead a DmStruct2<_,_>, ... -> DmStruct2 didn't work.


cannorin (approved):
Looks good to me!

PR #572, opened 2023-11-26
Matrix and Vector implement interfaces

gusty:
The original code had these implementations but we didn't include them because we needed #569


cannorin (approved):
Looks good!

gusty:
Note that I did the rebase in a separate commit intentionally in order to prove that #569 enabled this.

PR #573, opened 2023-11-26
+ .net8 target framework and F# language version 8

wallymathieu:
- Add: .net8 target framework for F#+ library
- Update: F# language version to 8
- Fix: A unique overload for method 'DateTime' could not be determined
- Fix: Indentation warnings that has turned into compilation error
- Fix: Fable 3.7 runs with downgraded .net versions and target framework


gusty (approved):
Looks good to me

PR #575, opened 2023-12-11
Limited applicative support for dictionaries

gusty:
Although dictionaries don't support the `Return` operation, we can go quite far with applicatives.

This is a good alternative to #448 instead of fighting the absence of `Return` we embrace it and add other missing methods in order to be able to use some applicative operations, including the applicative CE.

### DIFF
--- a/src/FSharpPlus/Extensions/Dict.fs
+++ b/src/FSharpPlus/Extensions/Dict.fs
@@ -73,6 +73,23 @@ module Dict =
             | None    -> ()
         dct :> IDictionary<'Key, 'U>
 
+    /// <summary>Combines values from three dictionaries using mapping function.</summary>
+    /// <remarks>Keys that are not present on every dictionary are dropped.</remarks>
+    /// <param name="mapping">The mapping function.</param>
+    /// <param name="source1">First input dictionary.</param>
+    /// <param name="source2">Second input dictionary.</param>
+    /// <param name="source3">Third input dictionary.</param>
+    ///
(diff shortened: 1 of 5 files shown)


src/FSharpPlus/Extensions/Dict.fs
```diff
@@ -73,6 +73,23 @@ module Dict =
...
+    /// <remarks>Keys that are not present on every dictionary are dropped.</remarks>
+    /// <param name="mapping">The mapping function.</param>
+    /// <param name="source1">First input dictionary.</param>
+    /// <param name="source2">Second input dictionary.</param>
+    /// <param name="source3">Third input dictionary.</param>
+    ///
+    /// <returns>The mapped dictionary.</returns>
+    let map3 mapping (source1: IDictionary<'Key, 'T1>) (source2: IDictionary<'Key, 'T2>) (source3: IDictionary<'Key, 'T3>) =
+        let dct = Dictionary<'Key, 'U> ()
+        let f = OptimizedClosures.FSharpFunc<_,_,_,_>.Adapt mapping
```
wallymathieu, line 86: It would be nice to have some high level docs about why we are supposed to use this method

wallymathieu (approved):

src/FSharpPlus/Extensions/Dict.fs
gusty: We can add it to the contributors or the design guideline.

PR #581, opened 2024-01-03
Implement (try)FindLastSliceIndex

### LINKED ISSUE #574: [API Proposal] String.(try)FindLastSliceIndex
We have String.{try}FindSliceIndex as an alternative to C# String.IndexOf(string), I could use also the equivalent of String.LastIndexOf(string).

```fs
module String;

    let findLastSliceIndex (slice: string) (source: string) =
        let index = source.LastIndexOf slice
        if index = -1 then
            ArgumentException("The specified substring was not found in the string.") |> raise
        else
            index
            
    let tryFindLastSliceIndex (slice: string) (source: string) =
        let index = source.LastIndexOf slice
        if index = -1 then None else Some index
```

I can put up a PR if approved.

rzikm:
Closes #574

I was not able to build the test project, so I rely on CI to run them.

### DIFF
--- a/src/FSharpPlus/Internals.fs
+++ b/src/FSharpPlus/Internals.fs
@@ -443,7 +443,7 @@ module FindLastSliceIndex =
     let sequenceEqual (a: _ seq) (b: _ seq) = a.SequenceEqual b
     #else
     let internal sequenceEqual (a: _ seq) (b: _ seq) :bool = Seq.compareWith Operators.compare a b = 0
-    module internal Q=
+    module internal Q =
         type queue<'a> = | Queue of 'a list * 'a list
 
         let empty = Queue([], [])
@@ -475,7 +475,7 @@ module FindLastSliceIndex =
         member _.Enqueue (v) = q <- Q.enqueue q v
         member _.Dequeue () =
             let (dequeued, next) = Q.dequeue q in q <- next
-            match dequeued with | Some v -> v | None -> failwith "Empty queue!"
+            match dequeued with | Some v -> v | None -> invalidOp "Empty queue!"
         member _.Count = Q.length q
     #endif
 
--- a/src/FSharpPlus/Internals.fs
+++ b/src/FSharpPlus/Internals.fs
@@ -479,6 +479,22 @@ module FindLastSliceIndex =
         member _.Count = Q.length q
     #endif
 
+    let listImpl (slice: _ list) (source: _ list) =
+        let cache = Queue<_>()
+        // List.length is O(n)
+        let sliceLength = slice.Length
+        let rec go last index source =
+            match source with
+            | h :: t ->
+                cache.Enqueue h
+                if cache.Count = sliceLength then
+                    let last = if sequenceEqual cache slice then index - sliceLength + 1 else last
+                    cache.Dequeue() |> ignore
+                    go last (index + 1) t
+                else go last (index + 1) t
+            | [] -> last
+        go -1 0 source
+
     let arrayImpl (slice: _ []) (source: _ []) =
         let revSlice = slice |> Array.rev
         let cache = Queue<_>()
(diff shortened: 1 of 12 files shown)


wallymathieu:
@rzikm Hi! What kind of issues around running the tests do you have?

gusty:
@rzikm That's fine, I also rely mostly in CIs run. You can keep adding commits, I will squash in the end.

Also, if you want to run them locally do it in Release mode, I don't think they will compile in DEBUG mode due to an old F# compiler bug.

gusty:
> I also got snagged because there is more than one WrappedListD type and I added a function to only one of them. Maybe there is a room for internal cleanup?

AFAIK There's only one, which is in General.fs line 70

The main purpose of these types was to test overloads paths, but @wallymathieu recently added a better mechanism for it in #557 

Still these wrapper types are sometimes useful to perform some tests as they are user-defined classes.

Independently from that, cleanups are always welcome, but then you have to consider either doing them in a separate PR or in a separate commit and then squash all the other unrelated commits.

My advice is a separate commit so you can draft it first and we can give early feedback.

BTW: when doing a separate PR or opening a new issue please mention me because somehow I never managed to get notifications for them until I participate.

rzikm:
> AFAIK There's only one, which is in General.fs line 70

The other one is in https://github.com/fsprojects/FSharpPlus/blob/54ffdcc9bf2638bca3d270e35c4e3de022ecc7d0/tests/FSharpPlusFable.Tests/FSharpTests/General/Util.fs#L67

gusty:
Oh, but those are Fable tests.
Fable tests are independent, they have their own set of wrappers.

We can unify some stuff, but sometimes it gets complicated.

gusty:
This seems to be ready to merge. The failure in appveyor is a transient one.

src/FSharpPlus/Internals.fs
```diff
@@ -419,6 +419,97 @@ module FindSliceIndex =
...
+                    let last = if cache.SequenceEqual slice then index - slice.Length + 1 else last
+                    cache.Dequeue() |> ignore
+                    go last (index + 1)
+                else go last (index + 1)
+            else last
+        go -1 0
+    let sequenceEqual (a: _ seq) (b: _ seq) = a.SequenceEqual b
+    #else
+    let internal sequenceEqual (a: _ seq) (b: _ seq) :bool = Seq.compareWith Operators.compare a b = 0
+    module internal Q=
```
gusty, line 446:
```suggestion
    module internal Q =
```

src/FSharpPlus/Internals.fs
```diff
@@ -419,6 +419,97 @@ module FindSliceIndex =
...
+    type Queue<'T> () =
+        let mutable q : Q.queue<'T> = Q.empty
+        interface IEnumerable<'T> with
+            member _.GetEnumerator () = let s = Q.toSeq q in s.GetEnumerator()
+        interface IEnumerable with
+            member _.GetEnumerator () = let s = Q.toSeq q in s.GetEnumerator() :> IEnumerator
+        member _.Enqueue (v) = q <- Q.enqueue q v
+        member _.Dequeue () =
+            let (dequeued, next) = Q.dequeue q in q <- next
+            match dequeued with | Some v -> v | None -> failwith "Empty queue!"
```
gusty, line 478:
```suggestion
            match dequeued with | Some v -> v | None -> invalidOp "Empty queue!"
```

gusty (changes requested):
Looks good to me. Just 2 nits.
In fact, there's room for optimizations by using ListCollector and possibly ArrayCollector, but to be fair, the same would apply to `findSliceIndex`, so it's better to leave both for a separate PR.

Thanks for this contribution !

gusty (approved):

PR #593, opened 2024-02-06
Ensure proper 'zip-applicative' behavior for async computations

gusty:
We expect that `computation.zip` accumulates errors, for async, task and valueTask.
While trying to assert it, we realized we need to fix some stuff, including some FSharp.Core functions.

### DIFF
--- a/src/FSharpPlus/Extensions/Async.fs
+++ b/src/FSharpPlus/Extensions/Async.fs
@@ -4,6 +4,23 @@ namespace FSharpPlus
 [<RequireQualifiedAccess>]
 module Async =
 
+    open System
+    open System.Threading.Tasks
+    
+    #if !FABLE_COMPILER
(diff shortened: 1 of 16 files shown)


src/FSharpPlus/Extensions/Async.fs
```diff
@@ -31,62 +34,61 @@ module Async =
...
+    /// <param name="mapper">The mapping function.</param>
+    /// <param name="async1">First async workflow.</param>
+    /// <param name="async2">Second async workflow.</param>
     #if FABLE_COMPILER
-    let map2 f x y = lift2 f x y
+    let map2 mapper (async1: Async<'T1>) (async2: Async<'T2>) : Async<'U> =
+        lift2 mapper async1 async2
     #else
-    let map2 f x y = async {
+    let map2 mapper (async1: Async<'T1>) (async2: Async<'T2>) : Async<'U> = async {
```
wallymathieu, line 48: a breaking change

Issue #578, opened 2023-12-18
Drop net45, net6 from F#+ 2 release


wallymathieu:
In order to simplify the release process, we should drop the target frameworks net45, net6 in the release. The main reason for this is that there is an [Issue with consuming DLLs built with .NET7 from .NET6](https://github.com/dotnet/fsharp/issues/14313).

Issue #577, opened 2023-12-17
Clean up preprocessor directives


wallymathieu:
We currently have NET7_0 directives in the code base, these should be NET7 or greater rather than a specific version.

```F#
#if NET7_0
```

See [preprocessor-directives](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/preprocessor-directives)

We have had trouble since NET?_OR_GREATER is not defined. Note the workaround in checking if not net45 et.c.. 

This is not a bug yet, since we have not yet released a version of F#+ compiled against a version greater than net6 (due to breaking changes in the F# compilation of assemblies).

See for instance:

```F#
#if !NET45 && !NETSTANDARD2_0 && !FABLE_COMPILER
```

PR #604, opened 2024-06-23
Bump to net8 and drop net45, fable3, netstandard

wallymathieu:
Drop net45 and bump to net8. Note that in the v1 branches net45 is still one of the target framework.

This is related to #578 and #577

### DIFF
--- a/src/FSharpPlus/Control/Applicative.fs
+++ b/src/FSharpPlus/Control/Applicative.fs
@@ -14,7 +14,7 @@ open FSharpPlus.Data
 type Apply =
     inherit Default1
 
-#if (!FABLE_COMPILER || FABLE_COMPILER_3) && !FABLE_COMPILER_4
+#if !FABLE_COMPILER
 
     static member        ``<*>`` (struct (f: Lazy<'T->'U>     , x: Lazy<'T>           )  , _output: Lazy<'U>             , [<Optional>]_mthd: Apply) = Lazy.apply f x                               : Lazy<'U>
     static member        ``<*>`` (struct (f: seq<_>           , x: seq<'T>            )  , _output: seq<'U>              , [<Optional>]_mthd: Apply) = Seq.apply  f x                               : seq<'U>
@@ -88,7 +88,7 @@ type Apply =
     static member inline InvokeOnInstance (f: '``Applicative<'T->'U>``) (x: '``Applicative<'T>``) : '``Applicative<'U>`` =
         ((^``Applicative<'T->'U>`` or ^``Applicative<'T>`` or ^``Applicative<'U>``) : (static member (<*>) : _*_ -> _) (f, x))
 
-#if (!FABLE_COMPILER || FABLE_COMPILER_3) && !FABLE_COMPILER_4
+#if !FABLE_COMPILER
 
 type Apply with
     static member inline ``<*>`` (struct (f: '``Monad<'T->'U>``      , x: '``Monad<'T>``     ) , _output: '``Monad<'U>``      , [<Optional>]_mthd:Default2) : '``Monad<'U>``       = Bind.InvokeOnInstance f (fun (x1: 'T->'U) -> Bind.InvokeOnInstance x (fun x2 -> Return.InvokeOnInstance (x1 x2)))
--- a/src/FSharpPlus/Control/Applicative.fs
+++ b/src/FSharpPlus/Control/Applicative.fs
@@ -28,7 +28,7 @@ type Apply =
     #if !FABLE_COMPILER
     static member        ``<*>`` (struct (f: Task<_>          , x: Task<'T>             ), _output: Task<'U>             , [<Optional>]_mthd: Apply) = Task.apply   f x : Task<'U>
     #endif
-    #if !NETSTANDARD2_0 && !FABLE_COMPILER
+    #if !FABLE_COMPILER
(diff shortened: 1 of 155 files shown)


src/FSharpPlus/Control/Applicative.fs
```diff
@@ -28,7 +28,7 @@ type Apply =
     #if !FABLE_COMPILER
     static member        ``<*>`` (struct (f: Task<_>          , x: Task<'T>             ), _output: Task<'U>             , [<Optional>]_mthd: Apply) = Task.apply   f x : Task<'U>
     #endif
-    #if !NET45 && !NETSTANDARD2_0 && !FABLE_COMPILER
+    #if !NETSTANDARD2_0 && !FABLE_COMPILER
```
gusty, line 31: Why do we keep netstandard ?

src/FSharpPlus/Control/Converter.fs
```diff
@@ -199,7 +199,7 @@ type Parse with
     
     static member inline Parse (_: ^R                  , _: Default2) : string -> ^R  = Parse.InvokeOnInstance
 
-    #if NET7_0
+    #if NET7_0_OR_GREATER
```
gusty, line 202: Shouldn't be net8 or greater ?

gusty (commented):
Shouldn't we remove old Fable versions as well ? Or is that for another PR?
Is TypeLevel missing here?

src/FSharpPlus/Control/Applicative.fs
wallymathieu: We unfortunately need to keep netstandard in order to avoid compilation error for type level. Perhaps we can get some help from @cannorin  on this?

src/FSharpPlus/Control/Converter.fs
wallymathieu: It should

wallymathieu:
> Shouldn't we remove old Fable versions as well ? Or is that for another PR? Is TypeLevel missing here?

We should remove the old fable versions as well.

src/FSharpPlus/Control/Applicative.fs
```diff
@@ -28,7 +28,7 @@ type Apply =
     #if !FABLE_COMPILER
     static member        ``<*>`` (struct (f: Task<_>          , x: Task<'T>             ), _output: Task<'U>             , [<Optional>]_mthd: Apply) = Task.apply   f x : Task<'U>
     #endif
-    #if !NET45 && !NETSTANDARD2_0 && !FABLE_COMPILER
+    #if !FABLE_COMPILER
```
wallymathieu, line 31: Is this needed since most of the file is excluded for Fable?

src/FSharpPlus/Control/Functor.fs
```diff
@@ -62,13 +62,13 @@ type Iterate =
...
 
-#if (!FABLE_COMPILER || FABLE_COMPILER_3) && !FABLE_COMPILER_4
+#if !FABLE_COMPILER
 
     static member Map ((x: Lazy<_>             , f: 'T->'U), _mthd: Map) = Lazy.map f x
     #if !FABLE_COMPILER
     static member Map ((x: Task<'T>            , f: 'T->'U), _mthd: Map) = Task.map f x : Task<'U>
     #endif
-    #if !NET45 && !NETSTANDARD2_0 && !FABLE_COMPILER
+    #if !FABLE_COMPILER
```
wallymathieu, line 71: Is this needed since it is in a block that is already excluded from Fable?

src/FSharpPlus/FSharpPlus.fsproj
```diff
@@ -19,16 +19,15 @@
     <GenerateAssemblyConfigurationAttribute>false</GenerateAssemblyConfigurationAttribute>
     <GenerateAssemblyFileVersionAttribute>false</GenerateAssemblyFileVersionAttribute>
     <AutoGenerateBindingRedirects>true</AutoGenerateBindingRedirects>
-    <Configurations>Debug;Release;Fable;Fable3;Test</Configurations>
+    <Configurations>Debug;Release;Fable;Test</Configurations>
     <Platforms>AnyCPU</Platforms>
     <LangVersion>8.0</LangVersion>
-    <LangVersion Condition=" '$(Configuration)' == 'Fable' OR '$(Configuration)' == 'Fable3' ">6.0</LangVersion>
+    <LangVersion Condition=" '$(Configuration)' == 'Fable' ">6.0</LangVersion>
```
wallymathieu, line 25: What should the language level for Fable4 be?

src/FSharpPlus/Control/Functor.fs
gusty: We can try removing it, once everything is green I guess.

src/FSharpPlus/FSharpPlus.fsproj
gusty: I think we decided to drop support for Fable 3, but let's do it in another PR.

wallymathieu: I already did drop it in this PR. Note the removal of a lot of ifdefs...

wallymathieu: We can cherry-pick those commits out of this PR though.

gusty: No, don't worry too much on that.

wallymathieu:
Many thanks @cannorin for the help!

PR #607, opened 2024-09-19
Monad instance for `Vector` and `Matrix`

cannorin:
This also makes `Vector<_, 'n>` and `Matrix<_, 'm, 'n>` monads.

ref: https://stackoverflow.com/questions/5802628/monad-instance-of-a-number-parameterised-vector


cannorin:
It looks like we must add `>>=` too since `>>=` is not automatically derived from `join` (but the converse is)

gusty (changes requested):
Looks great, do you mind adding a basic test ?

gusty (approved):
Cool, that's more than just basic tests, great job as always !

PR #610, opened 2024-10-17
Add NonEmptyList and NonEmptySeq Helper Functions

devinlyons:
This PR adds helper functions to the `NonEmptyList` and `NonEmptySeq` modules that bring them up to feature parity with the `List` and `Seq` modules.


wallymathieu:
Perhaps the doc-xml is either malformed or requires an update of fsdocs

```
  System.Xml.XmlException: The 'c' start tag on line 5987 position 77 does not match the end tag of 'summary'. Line 5987, position 143.
```
From https://github.com/fsprojects/FSharpPlus/actions/runs/11390820294/job/31734854875?pr=610

src/FSharpPlus/Data/NonEmptyList.fs
```diff
@@ -156,16 +195,103 @@ module NonEmptyList =
...
     /// <returns>The resulting average.</returns>
-    let inline average (list: NonEmptyList<'T>) = List.average (list.Head :: list.Tail)
+    let inline average (list: NonEmptyList<'T>) = Seq.average list
 
     /// <summary>Returns the average of the elements generated by applying the function to each element of the list.</summary>
     /// <param name="projection">The function to transform the list elements into the type to be averaged.</param>
     /// <param name="list">The input list.</param>
     /// <returns>The resulting average.</returns>
-    let inline averageBy (projection: 'T -> ^U) list = List.averageBy projection (list.Head :: list.Tail)
+    let inline averageBy (projection: 'T -> ^U)  (list: NonEmptyList<'T>) = List.averageBy projection (list.Head :: list.Tail)
```
cannorin, line 221: `NonEmptyList.average` uses `Seq.average` while `NonEmptyList.averageBy` uses `List.averageBy`, which seems inconsistent to me (I'm not very sure which is more performant, btw)

src/FSharpPlus/Data/NonEmptyList.fs
```diff
@@ -174,24 +300,268 @@ module NonEmptyList =
...
+    /// <returns>A list containing only the elements that satisfy the predicate.</returns>
+    let inline filter (predicate: 'T -> bool) (list: NonEmptyList<'T>): NonEmptyList<'T> = 
+      list |> toList |> List.filter predicate |> ofList
+
+    /// <summary>Returns a new collection containing only the elements of the collection for which the given predicate returns "true."</summary>
+    /// <param name="predicate">The function to test the input elements.</param>
+    /// <param name="list">The input list.</param>
+    /// <returns>A list containing only the elements that satisfy the predicate.</returns>
+    let inline tryFilter (predicate: 'T -> bool) (list: NonEmptyList<'T>): 'T list = 
+      list |> toList |> List.filter predicate
```
cannorin, line 351: Filter-like APIs would need some design discussions.

For example, I would personally prefer `tryFilter` to return `NonEmptyList<'T> option` rather than `'T list`. If I want a `'T list`, I can always write `xs |> NonEmptyList.toList |> List.filter`.

src/FSharpPlus/Data/NonEmptyList.fs
```diff
@@ -215,20 +585,298 @@ module NonEmptyList =
...
+    let pairwise (list: NonEmptyList<'T>) : NonEmptyList<'T * 'T> = 
+        Seq.pairwise list |> ofSeq
+
+    /// <summary>Splits the list into two lists, containing the elements for which the given function returns
+    /// <c>true</c> and <c>false</c> respectively.</summary>
+    /// <param name="predicate">A function to test each element of the list.</param>
+    /// <param name="list">The input list.</param>
+    /// <returns>A tuple containing the two lists.</returns>
+    let partition (predicate: 'T -> bool) (list: NonEmptyList<'T>) : NonEmptyList<'T> * NonEmptyList<'T> = 
+        list |> toList |> List.partition predicate |> fun (a, b) -> (ofList a, ofList b)
```
cannorin, line 600: Running `partition` on a non-empty list doesn't always return two non-empty lists?

src/FSharpPlus/Data/NonEmptyList.fs
```diff
@@ -215,20 +585,298 @@ module NonEmptyList =
...
+    let removeAt (index: int) (list: NonEmptyList<'T>) : NonEmptyList<'T> = 
+        Seq.removeAt index list |> ofSeq
+    
+    /// <summary>Removes multiple elements starting at the specified index.</summary>
+    /// <param name="index">The index at which to start removing elements.</param>
+    /// <param name="count">The number of elements to remove.</param>
+    /// <param name="list">The input list.</param>
+    /// <returns>The result list.</returns>
+    let removeManyAt (index: int) (count: int) (list: NonEmptyList<'T>) : NonEmptyList<'T> = 
+        Seq.removeManyAt index count list |> ofSeq
```
cannorin, line 652: Removing elements can produce empty lists?

src/FSharpPlus/Data/NonEmptyList.fs
```diff
@@ -215,20 +585,298 @@ module NonEmptyList =
...
+    /// <returns>The result list.</returns>
+    let skip (count: int) (list: NonEmptyList<'T>) : NonEmptyList<'T> = 
+        Seq.skip count list |> ofSeq
+    
+    /// <summary>Returns a list that skips elements while the predicate is true.</summary>
+    /// <param name="predicate">A function to test each element of the list.</param>
+    /// <param name="list">The input list.</param>
+    /// <returns>The result list.</returns>
+    let skipWhile (predicate: 'T -> bool) (list: NonEmptyList<'T>) : NonEmptyList<'T> = 
+        Seq.skipWhile predicate list |> ofSeq
```
cannorin, line 695: Skip would produce empty lists?

src/FSharpPlus/Data/NonEmptyList.fs
```diff
@@ -215,20 +585,298 @@ module NonEmptyList =
...
+    /// <returns>A tuple containing the two lists.</returns>
+    let splitAt (index: int) (list: NonEmptyList<'T>) : NonEmptyList<'T> * NonEmptyList<'T> = 
+        list |> toList |> List.splitAt index |> fun (a, b) -> (ofList a, ofList b)
+    
+    /// <summary>Splits the list into the specified number of lists.</summary>
+    /// <param name="count">The number of lists to create.</param>
+    /// <param name="list">The input list.</param>
+    /// <returns>A list of lists.</returns>
+    let splitInto (count: int) (list: NonEmptyList<'T>) : NonEmptyList<NonEmptyList<'T>> = 
+        Seq.splitInto count list |> Seq.map ofSeq |> ofSeq
```
cannorin, line 742: Split may produce empty lists?

src/FSharpPlus/Data/NonEmptyList.fs
```diff
@@ -215,20 +585,298 @@ module NonEmptyList =
...
+    /// <returns>The result list.</returns>
+    let takeWhile (predicate: 'T -> bool) (list: NonEmptyList<'T>) : NonEmptyList<'T> = 
+        Seq.takeWhile predicate list |> ofSeq
+    
+    /// <summary>Truncates the list to the specified length.</summary>
+    /// <param name="count">The maximum number of elements to include in the list.</param>
+    /// <param name="list">The input list.</param>
+    /// <returns>The truncated list.</returns>
+    let truncate (count: int) (list: NonEmptyList<'T>) : NonEmptyList<'T> = 
+        Seq.truncate count list |> ofSeq
```
cannorin, line 776: These may also produce empty lists. For `take` and `truncate`, we could check if `count` is positive.

src/FSharpPlus/Data/NonEmptyList.fs
```diff
@@ -215,20 +585,298 @@ module NonEmptyList =
...
+    /// <returns>The result list.</returns>
+    let updateAt (index: int) (value: 'T) (list: NonEmptyList<'T>) : NonEmptyList<'T> = 
+        Seq.updateAt index value list |> ofSeq
+    
+    /// <summary>Returns a list that contains the elements of the list for which the given function returns <c>true</c>.</summary>
+    /// <param name="predicate">A function to test each element of the list.</param>
+    /// <param name="list">The input list.</param>
+    /// <returns>The result list.</returns>
+    let where (predicate: 'T -> bool) (list: NonEmptyList<'T>) : NonEmptyList<'T> = 
+        Seq.where predicate list |> ofSeq
```
cannorin, line 858: (Filter-like API)

src/FSharpPlus/Data/NonEmptyList.fs
```diff
@@ -174,24 +300,268 @@ module NonEmptyList =
...
+    /// <returns>The only element of the list.</returns>
+    /// <exception cref="System.ArgumentException">Thrown when the input does not have precisely one element.</exception>
+    let inline exactlyOne (list: NonEmptyList<'T>) = Seq.exactlyOne list
+
+    /// <summary>Returns a new list with the distinct elements of the input list which do not appear in the itemsToExclude sequence, using generic hash and equality comparisons to compare values.</summary>
+    /// <param name="itemsToExclude">The sequence of items to exclude from the input list.</param>
+    /// <param name="list">The input list.</param>
+    /// <returns>A list that contains the distinct elements of list that do not appear in itemsToExclude.</returns>
+    /// <exception cref="System.ArgumentException">Thrown when itemsToExclude is null.</exception>
+    let inline except (itemsToExclude: #seq<'T>) (list: NonEmptyList<'T>) = Seq.except itemsToExclude list |> ofSeq
```
cannorin, line 322: (Filter-like API)

src/FSharpPlus/Data/NonEmptyList.fs
```diff
@@ -156,16 +195,103 @@ module NonEmptyList =
...
+    /// <returns>The resulting list comprising the values v where the chooser function returned Some(x).</returns>
+    let inline choose chooser (list: NonEmptyList<'T>) = list |> Seq.choose chooser |> ofSeq
+
+    /// <summary>
+    /// Applies a function to each element in a list and then returns a list of values v where the applied function returned Some(v).
+    /// </summary>
+    /// <param name="chooser">The function to be applied to the list elements.</param>
+    /// <param name="list">The input list.</param>
+    /// <returns>The resulting list comprising the values v where the chooser function returned Some(x).</returns>
+    let inline tryChoose chooser (list: NonEmptyList<'T>) = list |> Seq.choose chooser |> List.ofSeq
```
cannorin, line 237: (Filter-like API)

src/FSharpPlus/Data/NonEmptyList.fs
```diff
@@ -156,16 +195,103 @@ module NonEmptyList =
...
+    /// <param name="list">The input list.</param>
+    /// <returns>The concatenation of the transformed sublists.</returns>
+    let inline collect mapping (list: NonEmptyList<'T>) = list |> Seq.collect mapping |> ofSeq
+
+    /// <summary>For each element of the list, applies the given function.
+    /// Concatenates all the results and returns the combined list.</summary>
+    /// <param name="mapping">The function to transform each input element into a sublist to be concatenated.</param>
+    /// <param name="list">The input list.</param>
+    /// <returns>The concatenation of the transformed sublists.</returns>
+    let inline tryCollect mapping (list: NonEmptyList<'T>) = list |> Seq.collect mapping |> List.ofSeq
```
cannorin, line 259: `collect` should match the implementation of `>>=`?

devinlyons:
It looks like the documentation generator doesn't work with the `<c>text</c>` tags. I'll remove them.

src/FSharpPlus/Data/NonEmptyList.fs
devinlyons: `averageBy` was already implemented, so I mostly left it alone. I'll update it to match. 

I'm not sure which is more performant either. However, they should both by O(n).

src/FSharpPlus/Data/NonEmptyList.fs
devinlyons: I almost did it that way. I thought it made more sense to return `list` because both `list` and `option` have equivalent pattern-matching options (`[]` ~ `None`, `x :: xs` ~ `Some`) with `list` being more flexible and both provide the necessary type safety. However, `list` provided a more sane default for situations like `list |> NonEmptyList.tryFilter fn |> List.sort`. Otherwise, you would be left with something like `list |> NonEmptyList.tryFilter fn |> Option.map NonEmptyList.toList |> Option.defaultValue [] |> List.sort`.

src/FSharpPlus/Data/NonEmptyList.fs
devinlyons: Good point, I will add some checking, raise an exception, and provide a `tryPartition` function.

src/FSharpPlus/Data/NonEmptyList.fs
devinlyons: What if I changed the names to remove the expectation that an option be returned? Something like `filter` which returns a `list` and `filterUnsafe` which returns a `NonEmptyList`? Or perhaps `filter` and `attemptFilter` or something similar.

src/FSharpPlus/Data/NonEmptyList.fs
```diff
@@ -156,16 +195,103 @@ module NonEmptyList =
...
+    /// <param name="list">The input list.</param>
+    /// <returns>The resulting list comprising the values v where the chooser function returned Some(x).</returns>
+    let inline tryChoose chooser (list: NonEmptyList<'T>) = list |> Seq.choose chooser |> List.ofSeq
+
+    /// <summary>Divides the input list into lists (chunks) of size at most chunkSize.
+    /// Returns a new list containing the generated lists (chunks) as its elements.</summary>
+    /// <param name="chunkSize">The maximum size of each chunk.</param>
+    /// <param name="list">The input list.</param>
+    /// <returns>The list divided into chunks.</returns>
+    let inline chunkBySize chunkSize (list: NonEmptyList<'T>) =
```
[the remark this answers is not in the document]
devinlyons: `List.chunkBySize` will check that the `chunkSize` is greater than 0, but I will correct the return type.

src/FSharpPlus/Data/NonEmptyList.fs
cannorin: Again, if you want a sorted `'a list` instead of a sorted `NonEmptyList<'a>`, you can always convert it to `'a list ` first:
```fs
list |> NonEmptyList.toList |> List.filter fn |> List.sort
```

I believe the point of having dedicate `sort/trySort` functions in the `NonEmptyList` is to obtain `NonEmptyList<'a>` as a result.

Note that we do have a generic map function/operator and a generic `monad` computation expression, so if `NonEmptyList.tryFilter` returns `NonEmptyList<'a>`, you can write like these to obtain a filtered and sorted non-empty list:

```fsharp
xs |> NonEmptyList.tryFilter fn |> map NonEmptyList.sort

xs |> NonEmptyList.tryFilter fn |>> NonEmptyList.sort

monad {
  let! ys = xs |> NonEmptyList.tryFilter fn
  return ys |> NonEmptyList.sort
}
```

I strongly believe that `tryFoo` functions should have an `'a option` output, and it's for a good reason: it composes well with the other `tryFoo` functions when used within the computaiton expression:

```fsharp
monad {
  let! xs = tryParse "1,2,3" >>= NonEmptyList.tryOfSeq
  let! ys = trySomeOperation xs
  let! zs = tryAnotherOperation xs
  return! tryYetAnotherOperation ys zs
}
```

As we aim to provide some kind of a standard library, we must consider if the API design provides good ergonomics when used with arbitrary other APIs.

devinlyons: Okay. I'll make some updates.

gusty (approved):
Nice job @devinlyons and excellent review by @cannorin 

If you want to do more PRs like this, I have somewhere a small lists of collection missing functions from similar collections.

PR #611, opened 2024-10-29
Removed String.startsWith !FABLE_COMPILER condition

laurentpayot:
Hi, I had an error with `String.endsWith` when compiling a project using FSharpPlus with Fable. `String.endsWith` was not surrounded by a `!FABLE_COMPILER` symbol condition.
A Fable maintainer [fixed my issue by making `String.endsWith` work with Fable, _as well as `String.startsWith`_](https://github.com/fable-compiler/Fable/issues/3934#issuecomment-2429149490).

This PR simply removes the `!FABLE_COMPILER` symbol condition around `String.startsWith` that is not needed anymore.

### DIFF
--- a/.config/dotnet-tools.json
+++ b/.config/dotnet-tools.json
@@ -3,7 +3,7 @@
   "isRoot": true,
   "tools": {
     "fable": {
-      "version": "4.5.0",
+      "version": "4.23.0",
       "commands": [
         "fable"
       ]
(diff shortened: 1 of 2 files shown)


wallymathieu:
Do we need to update fable as well?

laurentpayot:
@wallymathieu Oopsie. I just bumped Fable to [version 4.23.0](https://github.com/fable-compiler/Fable/releases/tag/4.23.0) so the fix is uncluded.

wallymathieu (approved):

PR #614, opened 2025-02-02
Use F#10 for the main lib

gusty:
With the introduction of F#9 there seems to be a breaking change in backwards compatibility.

Although we can't fix the compiler, we can produce a lib that will work when consumed from F#9, still targeting .net8


.github/workflows/dotnetcore.yml
```diff
@@ -22,6 +22,8 @@ jobs:
       uses: actions/setup-dotnet@v3
       with:
         dotnet-version: |
+          10.0.x
+          9.0.x
```
wallymathieu: do we need .net 9 ?

gusty: Probably not

T-Gro:
@wallymathieu , @gusty : 

As part of fixing compiler regressions spotted in this PR, I was automatically running all targets and when successfully running them, I realized it also made an unintended push to gh-pages.

Would you mind having a look and reverting if needed?

gusty:
> Would you mind having a look and reverting if needed?

No worries, I just reverted. Thanks !

bartelink:
Logged https://github.com/fsprojects/FSharpPlus/issues/681 re the SRTP/memoization issue (which still persists with the .NET 11 toolchain).

~It's unclear to me at this point whether using the `net8.0` nupkg but compiling the tests on .NET 10/11 will repro; my main aim is to unblock https://github.com/dotnet/fsharp/pull/19785~ resolved by https://github.com/dotnet/fsharp/pull/20247

PR #623, opened 2025-06-05
Improve internal type signature

fxdmhtt:
The signature of GroupBy.Invoke function is abnormal, which causes the secondary encapsulated '``Collection<'T>`` type to be incorrectly identified (identified as Id<'T>).

Refer to ChunkBy.Invoke, this should be just a typo.


src/FSharpPlus/Control/Collection.fs
```diff
@@ -305,7 +305,7 @@ type GroupBy =
     static member GroupBy (x: list<'T>, f: 'T->'Key, _: list<'Key*list<'T>>, [<Optional>]_impl: GroupBy) = Seq.groupBy f x |> Seq.map (fun (x, y) -> x, Seq.toList  y) |> Seq.toList
     static member GroupBy (x: 'T []   , f: 'T->'Key, _: ('Key*('T [])) []  , [<Optional>]_impl: GroupBy) = Seq.groupBy f x |> Seq.map (fun (x, y) -> x, Seq.toArray y) |> Seq.toArray
 
-    static member inline Invoke    (projection: 'T->'Key) (source: '``C<'T>``) : '``C<'Key * 'C<'T>>`` = 
+    static member inline Invoke    (projection: 'T->'Key) (source: '``Collection<'T>``) : '``Collection<'Key * 'Collection<'T>>`` =
```
wallymathieu, line 308: Note the ticks in the type signature. The C here is more for documentation rather than an actual type

wallymathieu: For sure, I’m wondering if @fxdmhtt has some issue with code? Something that does not compile?

fxdmhtt:
I have seen your reply and I completely agree with what you said.

I think it is more of my own problem and I didn't want to bother you about it.

Please allow me to extract a minimal reproducible code tomorrow. Thank you very much!

fxdmhtt:
Thank you very much.

The minimal example code is as follows:

```F#
open FSharpPlus

let inline groupby (key: 'T -> 'Key when 'Key : equality) (sequence: '``Collection<'T>``) : Map<'Key, '``Collection<'T>``> =
    sequence
    |> groupBy key
    |> Map.ofSeq
```

I will get a wavy line error on groupBy:
The type 'Internals.Id<'Key * Internals.Id<'T>>' is not compatible with the type '('Key * 'Collection<'T>) seq'.

This code doesn't really mean anything, I'm just more familiar with Python so I'm trying to simulate the behavior of the Toolz package. This is a toy.

Specifically, I'm trying to wrap the generic groupBy as [Toolz.groupby](https://toolz.readthedocs.io/en/latest/api.html#toolz.itertoolz.groupby)

I'm a beginner in F# and don't know if this problem is caused by FSharpPlus, but as of now, I really don't have any idea to solve this compilation error.

OS: Windows 10
.NET: 9.0

Thanks again!

gusty (changes requested):
See in `Distinct`, `DistinctBy`, `Rev`, and all ` Sort` related Invokables have also the abbreviated `C<`T>` type signature.

Please apply the same change to them, so we can merge this.

fxdmhtt:
> > There are many `C<'T>` in the code, but the only lonely `C<'T>` is in the GroupBy function. I don't know which "another" you are referring to.
> 
> See my review, or just do a find ` C<'T>`.
> 
> > Have you considered replacing all `C<'T>` in batches?
> 
> I rarely do these kind of PRs that only impact code readability, I'm not saying it's bad, it's just that since I have limited time I prefer to focus myself in solving functionality issues. So, it's good to have collaborators who take care of code uniformity. It definitely helps in the long run.

Please review the patch

gusty (approved):
OK from my side, let's see if CI agree as well ;)

fxdmhtt:
Sorry. I saw the CI error, but I don't have the ability to fix it.

In fact, I raised this issue because I encountered difficulties in using F# type inference.

I need your help.

Thank you very much.

gusty:
No worries. I will fix it and merge it. 
Thanks !

gusty:
Yes, whatever is inside the backticks is not being parsed, so no matter what you write there, it won't have any different effect.
Of course, unless it matches or differ from existing ones, which was the case here.

wallymathieu:
@fxdmhtt thanks for the contribution! 👍

PR #632, opened 2025-06-10
Fix Traverse for Async seqs

gusty:
The problem with the current implementation is that it ignores overloads defined in the custom IEnumerable.

Moreover, in these cases the resulting type is wrong:

```
> let b = traverse async.Return (MyCustomIEnumerable.ofSeq (Seq.initInfinite id))  ;;
val b: Async<int seq>   ===> it should be Async<int MyCustomIEnumerable>
```


gusty:
This was reproduced here 917b669bd9e9a6e41b4fa7757c90fc747e8ef3d8 and fixed at ba1d4ff8b77bdab6599945b7ffa5f853839991b5

PR #633, opened 2025-06-10
Fix internal Id type being selected in overload resolution

gusty:
This was described here #623

### DIFF
--- a/tests/FSharpPlus.Tests/Collections.fs
+++ b/tests/FSharpPlus.Tests/Collections.fs
@@ -291,3 +291,10 @@ module Collections =
         
         let m = choose Some ((ofSeq :seq<_*_> -> Map<_,_>) (seq ["a", 1; "b", 2]))
         Assert.IsInstanceOf<Option<Map<string,int>>> (Some m)
+
+    // Compile tests
+    
+    let inline mapOfGroup (key: 'T -> 'Key when 'Key : equality) (sequence: '``Collection<'T>``) : Map<'Key, '``Collection<'T>``> =
+        sequence
+        |> groupBy key
+        |> Map.ofSeq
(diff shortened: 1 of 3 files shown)


wallymathieu (approved):

tests/FSharpPlus.Tests/Collections.fs
```diff
@@ -291,3 +291,10 @@ module Collections =
         
         let m = choose Some ((ofSeq :seq<_*_> -> Map<_,_>) (seq ["a", 1; "b", 2]))
         Assert.IsInstanceOf<Option<Map<string,int>>> (Some m)
+
+    // Compile tests
+    
+    let inline mapOfGroup (key: 'T -> 'Key when 'Key : equality) (sequence: '``Collection<'T>``) : Map<'Key, '``Collection<'T>``> =
```
gusty, line 297: @fxdmhtt  this function is kind of useful.

I already thought about grouping in the past and realized that ideally they should return non-empty collections.
Returning a map makes sense for the eager `groupBy` (as opposed to the lazy `chunkBy`), so maybe we should create an issue to discuss a function that groups into a NonEmptyMap.

fxdmhtt: I don't understand how this works.

Why can I write the correct type signature by simply making the Id redundant?

fxdmhtt: In fact, when I was writing the secondary encapsulation library, I encountered many situations where I had to delete the type signature to compile. But I thought it was caused by my weak F# ability, so I didn't report it to you.

Can you tell me what happened?

fxdmhtt:
So the root cause is the defects of F# itself, and we just use some "magic" to circumvent these defects.

Then there is another problem: if the function signature does not have a complete type description, the API documentation and fsi files are not very standardized. If the function has a complete type description, in order to circumvent the defects of F# itself, it may be necessary to modify the underlying reference library, but this is generally impossible.

It seems that the ceiling of F# is a bit low.

fxdmhtt:
If F# is willing to support HKT, it will be closer to Haskell.

That is the classic question: Why didn't Microsoft directly introduce Haskell into .NET?

So I don't think I will see that day.

To be honest, I feel that Microsoft is letting F# run wild.

wallymathieu:
HKT would probably not give you Haskell on .net. There are many features in Haskell...

gusty:
Starting with the lazy evaluation by default.
But Haskell and HKTs are not the same thing. Other languages support HKTs, which are maybe closer to F#.

PR #638, opened 2025-06-15
Documenation about arrow

brainfucknow:


docsrc/content/abstraction-arrow.fsx
```diff
@@ -68,6 +68,146 @@ Rules
...
+Tutorial
+--------
+
+A value of type ``Arrow<'T, 'U>`` represents a process that takes as input a value of type ``'T`` and outputs a value of type ``'U``.
+
+Let's define a simple arrow type based on a function:
+
+*)
+
+#if INTERACTIVE
```
wallymathieu, line 80: Why not use the same pattern as the other files?

docsrc/content/abstraction-arrow.fsx
```diff
@@ -68,6 +70,148 @@ Rules
...
+
+let hOutput = runF h 8
+// Result: 29 (because (8/2) + (8*3+1) = 4 + 25 = 29)
+
+(**
+
+How does ``h`` work? The process is: ``split >>> first f >>> second g >>> unsplit (+)``
+
+When applied to ``8``:
+1. ``8`` → ``(8, 8)`` via ``split``
```
wallymathieu, line 153: I don't really see that the liftA2 uses split?

